PR #17, opened 2017-05-13
Add support for .net 4.7 and update documentation

kmcgain:
New dockerfile that downloads and installs 4.7. Updates documentation to refer to 4.7 as the latest version.

### DIFF
--- a/4.7/Dockerfile
+++ b/4.7/Dockerfile
@@ -1,0 +1,8 @@
+FROM microsoft/windowsservercore
+
+RUN powershell -NoProfile -Command \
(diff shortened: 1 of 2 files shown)


kendrahavens:
@kmcgain Thank you for this contribution. We are still working through build infrastructure changes that would allow the publishing of the 4.7 image. Until that is finished, we'll have to hold off on updating the README to 4.7. Ideally, 4.7 would ship rolled up in the windowsservercore image itself.

kmcgain:
@kendrahavens that makes sense. It would ideally be rolled into the windowsservercore image. Perhaps this should be left open for others to use as a temporary measure for now.

richlander:
Using #19 as the primary PR for this now.

PR #19, opened 2017-06-05
Add support for .NET Framework 4.7

richlander:
- In support of https://github.com/dotnet/designs/issues/7
- Builds on top of #17.
- Tried to convert the Dockerfiles to PowerShell. Was unable to after spending non-zero time futzing with it. Couldn't figure out how to do the combination of: run the setup, pass arguments to the setup and wait on the setup until deleting it. That all worked w/o issue in cmd.exe.

### DIFF
--- a/4.7/Dockerfile
+++ b/4.7/Dockerfile
@@ -1,0 +1,8 @@
+FROM microsoft/windowsservercore
+
+RUN powershell -NoProfile -Command \
+    Invoke-WebRequest -Uri "https://download.microsoft.com/download/D/D/3/DD35CC25-6E9C-484B-A746-C5BE0C923290/NDP47-KB3186497-x86-x64-AllOS-ENU.exe" \
+    -OutFile C:\NDP47-KB3186497-x86-x64-AllOS-ENU.exe
+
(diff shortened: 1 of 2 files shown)


4.7/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM microsoft/windowsservercore
+
+RUN powershell -NoProfile -Command \
+    Invoke-WebRequest -Uri "https://download.microsoft.com/download/D/D/3/DD35CC25-6E9C-484B-A746-C5BE0C923290/NDP47-KB3186497-x86-x64-AllOS-ENU.exe" \
+    -OutFile C:\NDP47-KB3186497-x86-x64-AllOS-ENU.exe
+
+RUN ["C:\\NDP47-KB3186497-x86-x64-AllOS-ENU.exe", "/q", "/norestart", "/ChainingPackage", "ADMINDEPLOYMENT"] \
```
MichaelSimons: Why multiple layers?  It seems inefficient that a layer is getting created that contains the .exe that is getting removed later.

richlander: Oops. I reduced the layers from the original PR but not this one. Not sure why I didn't do that. Thanks.

richlander:
Squashed my commits into one, leaving just the two commits. This means that the commit from @kmcgain will remain in-tact. Will use a merge commit when we land the PR.

PR #27, opened 2017-10-05
Update NGEN compilation

### LINKED ISSUE #25: Native images missing?
While investigating why a PowerShell healthcheck script consisting solely of `iwr http://localhost/health -usebasic` took three seconds and a full CPU core to execute, I discovered the .NET 4.7 container doesn't seem to have valid native images for framework assemblies.  Running `ngen update` from inside the container built new images from mscorlib up, and reduced the script time to about 200ms.

Is it possible to have these either built into the base image, or background generated on container start?

richlander:
- NGEN compilation was failing due to certain Windows dependencies not being enabled in Windows containers yet.
- The COMPLUS_NGenProtectedProcess_FeatureEnabled is used as a workaround.

Fixes #25


artisticcheese:
I see newer 2019 images do not have this optimization built in. Was it fixed in 2019?

PR #28, opened 2017-10-10
Add .NET Framework 4.7.1 image

richlander:
Waiting for .NET Framework 4.7.1 to ship.

Fixes #26

PR #30, opened 2017-10-18
Dotnet471

Sugol:


MichaelSimons:
@sugol - Thanks for the submission.  How is this different than #28 which was merged yesterday?

PR #36, opened 2017-10-26
Update readme for 10.0.14393.1770 release

MichaelSimons:
Looks like the readme wasn't updated when the 1770 images were published.


richlander:
I published the README on DH. Thanks!

PR #37, opened 2017-11-01
Introduce build Dockerfiles

### LINKED ISSUE #38: Add support for building .NET Framework applications with Docker
Microsoft should produce Docker images that can be utilized to build .NET Framework applications.

MichaelSimons:
This PR introduces Docker images which can be used to build .NET Framework projects with.  A new dotnet-framework-build repo will be created to host the new build related images.  This follows the pattern used in the ASP.NET Core Docker repos.

Remaining work:

~~- [ ] Create Dockerfiles capable of building any .NET Framework versioned application~~ Will be handled with a separate PR.
- [x] Utilize stand-alone targeting pack installer once defined
- [x] Finalize on tag - specifically the version differentiator.  We need the ability to rev these images outside of the .NET Runtime.

Would appreciate any feedback.  You can try out images I built and pushed to [msimons/dotnet-framework-build](https://hub.docker.com/r/msimons/dotnet-framework-build/)

Fixes #38


4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# escape=`
```
shirhatti: If we're introducing a `build` and `runtime` folder, I'd like to see that consistently applied to all Dockerfiles even those without a build image right now

README.dotnet-framework-build.md
```diff
@@ -0,0 +1,108 @@
+# Supported Windows Server 2016 Version 1709 (Fall Creators Update) amd64 tags
+
+- [`4.7.1-1-windowsservercore-1709`, `4.7.1-windowsservercore-1709` (*4.7.1-windowsservercore-1709/build/Dockerfile*)](https://github.com/Microsoft/dotnet-framework-docker/blob/master/4.7.1-windowsservercore-1709/build/Dockerfile)
```
shirhatti: I'm assuming for the time being, you've just copied the README as is?
All these tags are missing the '-build'

README.md
```diff
@@ -1,140 +1,9 @@
-# Supported Windows Server 2016 Version 1709 (Fall Creators Update) amd64 tags
+dotnet-framework-docker
```
shirhatti: Personally, I'm not a fan of the split READMEs. Especially given the fact that I envision a combined "How to use these images" with a multi-stage build.

4.7.1/build/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# escape=`
```
shirhatti: Any reason you had to change the escape character? If possible, I'd like to be consistent across the build and runtime Dockerfiles

4.7.1/build/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM microsoft/dotnet-framework:4.7.1-windowsservercore
+
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
+
+# Install Git
+RUN Invoke-WebRequest -UseBasicParsing https://github.com/git-for-windows/git/releases/download/v2.14.3.windows.1/MinGit-2.14.3-64-bit.zip -outfile git.zip; `
```
shirhatti: It's probably been discussed before, but why add git?

README.md
MichaelSimons: I should have called out that the plan is for two docker hub repos - `microsoft/dotnet-framework` and `microsoft/dotnet-framework-build`.  Because of this two READMEs are required - one per repo.  The patterns used here is following what is currently done in the ASP.NET Core.

I am certainly open to a discussion on the two vs one repo.

README.dotnet-framework-build.md
MichaelSimons: No the readme wasn't just copied.  A fair bit of it is the same but the sample and tags are different amongst other minor parts.

The `build` suffix is not necessary as the plan is for a new `microsoft/dotnet-framework-build` repo.

4.7.1/build/Dockerfile
MichaelSimons: `git` was included because it is a common ask.  I am open to feedback on whether or not it should be included.

4.7.1/build/Dockerfile
MichaelSimons: I had to switch it at one point when I was using a multi-stage build.  I ended up in a situation where double escaping of `\` was needed which was pretty ugly.  I didn't end up going down that approach so I could revert this.  I will point out however that Docker points out the usefulness of using ` in Windows Dockerfiles to avoid issues - see https://docs.docker.com/engine/reference/builder/#escape.  I agree consistency across all Dockerfiles is best.  What is your opinion on switching it even though it is not strictly necessary at this point?

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: I have no problems with that.

README.dotnet-framework-build.md
shirhatti: Ahh. My bad.
1 GH repo builds 2 DH repos?

MichaelSimons: correct

4.7.1/build/Dockerfile
MichaelSimons: I will note this is inconsistent with the .NET Core Windows Dockerfiles - perhaps we should remove it to begin with.

richlander: Scenario-wise, I think it makes sense to include git. I've used that in our .NET Core SDK images. It enables a certain style of deployment. Certainly, it only does that for one SCM, but I think that's OK. People are free to add their own, but git is what the team uses and many customers use, so it feels OK to include just it.

I looked at [buildpack-deps](https://hub.docker.com/_/buildpack-deps/). Nothing else popped out to me to add that makes sense on Windows. We could consider a component like [curl](https://curl.haxx.se/download.html). As you can see, Powershell provides analogous functionality. If we wanted to avoid setting a specific shell, then curl would be more attractive.

richlander:
What is the thinking on the builder image once 4.7.2 ships? Move it to 4.7.2 or keep it and 4.7.1? For a build scenario, it seems super attractive to have just one and make it capable of building all .NET versions.

richlander:
Is there a test FX that comes with the build tools? If not, how would someone add it?

We are talking about enabling the "dotnet" tools as a runtime-agnostic toolset. When we do that, We'll want to include them in this image. Yes?

chrisrpatterson:
The test framework tools should be included as part of the build sku as they are as important as MSBuild for those scenarios.

shirhatti:
> What is the thinking on the builder image once 4.7.2 ships? Move it to 4.7.2 or keep it and 4.7.1? For a build scenario, it seems super attractive to have just one and make it capable of building all .NET versions.

I'd rather have a different builder image for each framework version. We can re-organize the order in which we add dependencies so we have layer sharing up until the point of installing the targeting packs. I've seen nothing to indicate folks want a combined build image.

richlander:
@shirhatti A combined build image seems to align very closely with the ASP.NET Core kitchensink image.

I think the big question is why developers would want to pick 4.7.1 vs. 4.7.2. What is that reason for you?

4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    Remove-Item -Force git.zip; `
+    setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\git')
+
+# Install NuGet CLI
+RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; `
+    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v4.3.0/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe; `
+    setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\NuGet')
+
+# Install Targeting Pack
+RUN Invoke-WebRequest https://download.microsoft.com/download/9/0/1/901B684B-659E-4CBD-BEC8-B3F06967C2E7/NDP471-DevPack-ENU.exe -OutFile NDP462-DevPack-KB3151934-ENU.exe; `
```
chrisrpatterson: Not that is matters functionally but you should keep consistent file names after download.  Here you are downloading NDP471-DevPack-ENU.exe and calling it NDP462-DevPack-KB3151934-ENU.exe

richlander:
I just made PR #39. We're going to need to keep updating the contents of the repo. I suggest we limit this PR (or a new one) to just the builder Dockerfile. Once we've merged that, then I suggest we make all the other changes. This PR has the potential to hold the repo hostage while we get the builder plan in place. Cool?

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: This wasn't intentional - thanks for pointing it out - fixed.

4.7.1/build/Dockerfile
MichaelSimons: I think we should start off without including git.  I think the usage scenarios for it are limited.  I would rather be in the situation where we get requests to add it.  If we release with it, it will be difficult to remove it down the road.  Lastly, we don’t have it in the .NET Core SDK images and I don’t recall anyone asking for it.  I talked to @richlander about this and he is fine with excluding it to begin with.

MichaelSimons:
@shirhatti, @chrisrpatterson, @richlander - please take another look.  Per Rich's request, I reverted the readme restructuring and will handle with a followup PR.

4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -5,19 +5,19 @@ SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPref
 # Install NuGet CLI
 RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
     Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v4.3.0/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe; \
-    setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\NuGet')
+    & setx /M PATH $($Env:PATH + ';' + $Env:ProgramFiles + '\NuGet')
```
shirhatti: Why setx over `ENV` directive in the Dockerfile? I thought setx won't won't be inherited if another Dockerfile (aspnet-build) depends on this?

3.5/runtime/Dockerfile
```diff
@@ -1,11 +1,11 @@
...
-    \Windows\Microsoft.NET\Framework\v4.0.30319\ngen update & \
-    \Windows\Microsoft.NET\Framework64\v2.0.50727\ngen update & \
-    \Windows\Microsoft.NET\Framework\v2.0.50727\ngen update
+FROM microsoft/windowsservercore
+
+ADD install /build
+RUN DISM /Online /Add-Package /PackagePath:C:\build\microsoft-windows-netfx3-ondemand-package.cab & \
+    DISM /Online /Add-Package /PackagePath:C:\build\patch\Windows10.0-KB3213986-x64.cab  & \
+    rd /s /q C:\build
+RUN set COMPLUS_NGenProtectedProcess_FeatureEnabled=0 & \
```
shirhatti: Probably the wrong PR to raise this concern, but I just noticed that we aren't checking exit codes when chaining together exe invocations.
I'm not familiar with cmd, but the PowerShell equivalent would read something like this
```Dockerfile
RUN set COMPLUS_NGenProtectedProcess_FeatureEnabled=0; \
    if($LASTEXITCODE -ne 0){ Exit $LASTEXITCODE } ; \
    & 'C:\Windows\Microsoft.NET\Framework64\v4.0.30319\ngen' update ; \
    if($LASTEXITCODE -ne 0){ Exit $LASTEXITCODE } ; \
    & 'C:\Windows\Microsoft.NET\Framework\v4.0.30319\ngen' update ; \
    if($LASTEXITCODE -ne 0){ Exit $LASTEXITCODE } ; \
    & 'C:\Windows\Microsoft.NET\Framework64\v2.0.50727\ngen' update ; \
    if($LASTEXITCODE -ne 0){ Exit $LASTEXITCODE } ; \
    & 'C:\Windows\Microsoft.NET\Framework\v2.0.50727\ngen' update ;
```

MichaelSimons: @shirhatti - Can you log a separate issue for this?

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: `ENV` does not support path concatenation on Windows - see https://github.com/moby/moby/issues/22017.  My tests on with the aspnet-build images indicate this works as expected.  I will do some more testing though.

3.5/runtime/Dockerfile
shirhatti: Filed #43

MichaelSimons:
Couple discoveries of note:
1. Never set the Path variable multiple times within a windows Dockerfile.  The overhead of each layer is terrible - Read more about this problem at https://github.com/moby/moby/issues/22017
2. Need to remove the VS installer and cached VSIX packages as a result of installing VS Test Agent and VS Build Tools.  I'm working with the VS folks to figure out the correct way to do this.

MichaelSimons:
Got direction from VS on how to cleanup installer - pushed change.  

The following are the image layer sizes (495.3MB + Base runtime image)

```
PS C:\> docker history msimons/dotnet-framework-build:4.7.1-windowsservercore-1709
IMAGE               CREATED             CREATED BY                                      SIZE
f317254b1984        6 minutes ago       powershell -Command $ErrorActionPreference...   141MB
aa12aa4d5508        6 minutes ago       powershell -Command $ErrorActionPreference...   31.3MB
e76494ecd19a        7 minutes ago       powershell -Command $ErrorActionPreference...   290MB
d0757b16aa7f        27 hours ago        powershell -Command $ErrorActionPreference...   33MB
2f04cc2a4931        27 hours ago        powershell -Command $ErrorActionPreference...   41kB
5029f1cce4a2        2 days ago          cmd /S /C set COMPLUS_NGenProtectedProcess...   979MB
<missing>           2 weeks ago         Install update 10.0.16299.64                    962MB
<missing>           7 weeks ago         Apply image 10.0.16299.15                       4.62GB
```

richlander:
We decided to use zips instead of MSIs for this image. Each zip is named for the highest target framework version it contains, one per .NET Framework version series.

* https://dotnetbinaries.blob.core.windows.net/referenceassemblies/v4.0.zip
* https://dotnetbinaries.blob.core.windows.net/referenceassemblies/v4.5.2.zip
* https://dotnetbinaries.blob.core.windows.net/referenceassemblies/v4.6.2.zip
* https://dotnetbinaries.blob.core.windows.net/referenceassemblies/v4.7.1.zip

We intend to create one "everything included" image to start and then intend to create a more opt-in experience in 2018 that starts with zero targeting packs.

MichaelSimons:
After updating the images to include the targeting packs >= 4.0, the image layer sizes look like the following (838MB + Base image):

```
PS D:\repos\Sandbox\FullFxTestApp471> docker history msimons/dotnet-framework-build:4.7.1-windowsservercore-1709
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
17f27146e5e2        17 hours ago        powershell -Command $ErrorActionPreference...   434MB
<missing>           17 hours ago        powershell -Command $ErrorActionPreference...   29.3MB
<missing>           17 hours ago        powershell -Command $ErrorActionPreference...   342MB
<missing>           18 hours ago        powershell -Command $ErrorActionPreference...   33.1MB
<missing>           5 days ago          powershell -Command $ErrorActionPreference...   41kB
<missing>           7 days ago          cmd /S /C set COMPLUS_NGenProtectedProcess...   979MB
<missing>           2 weeks ago         Install update 10.0.16299.64                    962MB
<missing>           7 weeks ago         Apply image 10.0.16299.15                       4.62GB
```

@richlander - this is ready for final review.

PR #52, opened 2017-11-30
Add .NET Core SDK to Build images

### LINKED ISSUE #49: Build images don't support building SDK styled projects
The build images (e.g microsoft/dotnet-framework-build:4.7.1) don't support building SDK styled projects (e.g. the [Full Fx 4.7 sample Docker project](https://github.com/Microsoft/dotnet-framework-docker-samples/blob/master/dotnetapp-4.7/dotnetapp-4.7.csproj)).  In order to support these projects the .NET Core SDK would need to be added.

MichaelSimons:
Fixes #49 

I made a slash style change because I noticed the '/' wasn't getting inverted when setting the PATH.  I thought it was best to use std Windows '\' just in case something was reading the PATH and didn't handle '/'.

The SDK adds 360MB to the image.

### DIFF
--- a/4.7.1-windowsservercore-1709/build/Dockerfile
+++ b/4.7.1-windowsservercore-1709/build/Dockerfile
@@ -3,8 +3,9 @@ FROM microsoft/dotnet-framework:4.7.1-windowsservercore-1709
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
 # Install NuGet CLI
+ENV NUGET_VERSION 4.4.1
 RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
-    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v4.4.1/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
+    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v$Env:NUGET_VERSION/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
 
 # Install VS Test Agent
 RUN Invoke-WebRequest -UseBasicParsing https://download.visualstudio.microsoft.com/download/pr/100232757/8a386d27295953ee79281fd1f1832e2d/vs_TestAgent.exe -OutFile vs_TestAgent.exe; \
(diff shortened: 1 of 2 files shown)


4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -3,30 +3,34 @@ FROM microsoft/dotnet-framework:4.7.1-windowsservercore-1709
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
 # Install NuGet CLI
+ENV NUGET_VERSION 4.4.1
 RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
-    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v4.4.1/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
+    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v$Env:NUGET_VERSION/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
```
ravimeda: It is easy to miss the `v` in `/v$Env:NUGET_VERSION/`. How about including this character in the version variable as in `ENV NUGET_VERSION v4.4.1`? On the other hand, having a non-numeric in a version isn't better either.

4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -3,30 +3,34 @@ FROM microsoft/dotnet-framework:4.7.1-windowsservercore-1709
...
 # Install VS Test Agent
 RUN Invoke-WebRequest -UseBasicParsing https://download.visualstudio.microsoft.com/download/pr/100232757/8a386d27295953ee79281fd1f1832e2d/vs_TestAgent.exe -OutFile vs_TestAgent.exe; \
     Start-Process vs_TestAgent.exe -ArgumentList '--quiet', '--norestart', '--nocache' -NoNewWindow -Wait; \
     Remove-Item -Force vs_TestAgent.exe; \
 # Install VS Build Tools
     Invoke-WebRequest -UseBasicParsing https://download.visualstudio.microsoft.com/download/pr/100196686/e64d79b40219aea618ce2fe10ebd5f0d/vs_BuildTools.exe -OutFile vs_BuildTools.exe; \
-    Start-Process vs_BuildTools.exe -ArgumentList '--add', 'Microsoft.VisualStudio.Workload.MSBuildTools', '--quiet', '--norestart', '--nocache' -NoNewWindow -Wait; \
+    # Installer won't detect DOTNET_SKIP_FIRST_TIME_EXPERIENCE if ENV is used, must use setx /M
+    setx /M DOTNET_SKIP_FIRST_TIME_EXPERIENCE 1; \
+    Start-Process vs_BuildTools.exe -ArgumentList '--add', 'Microsoft.VisualStudio.Workload.MSBuildTools', '--add', 'Microsoft.VisualStudio.Workload.NetCoreBuildTools', '--quiet', '--norestart', '--nocache' -NoNewWindow -Wait; \
```
ravimeda: Slightly unrelated to the current changes, but a clarifying question - `vs_BuildTools.exe` does this write a log(s) or temp files? If so, are these removed in one of `Remove-Item` below?

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: I had thought about this but decided against it.  The version is really just 4.4.1.  The download url is what uses the 'v' prefix.  Since this is an env var within the image I felt it was better to use the real version number.

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: Yes it does - they are written to the %TEMP% dir which is cleaned up at https://github.com/Microsoft/dotnet-framework-docker/pull/52/files#diff-fb88ca50fb6c815bef8dbb9768e66146R21

4.7.1-windowsservercore-1709/build/Dockerfile
```diff
@@ -3,30 +3,34 @@ FROM microsoft/dotnet-framework:4.7.1-windowsservercore-1709
...
+    + \";${Env:ProgramFiles}\NuGet\" \
+    + \";${Env:ProgramFiles(x86)}\Microsoft Visual Studio\2017\TestAgent\Common7\IDE\CommonExtensions\Microsoft\TestWindow\" \
+    + \";${Env:ProgramFiles(x86)}\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin\")
 
 # Install Targeting Packs
 RUN @('4.0', '4.5.2', '4.6.2', '4.7.1') \
     | %{ \
         Invoke-WebRequest -UseBasicParsing https://dotnetbinaries.blob.core.windows.net/referenceassemblies/v${_}.zip -OutFile referenceassemblies.zip; \
-        Expand-Archive referenceassemblies.zip -DestinationPath \"${Env:ProgramFiles(x86)}/Reference Assemblies/Microsoft/Framework/.NETFramework\"; \
+        Expand-Archive referenceassemblies.zip -DestinationPath \"${Env:ProgramFiles(x86)}\Reference Assemblies\Microsoft\Framework\.NETFramework\"; \
```
ravimeda: Again unrelated to this change - I suggest using a '-Force' parameter. In a unlikely case, if a folder/file with the same name exists, then it is overwritten without waiting for confirmation.

4.7.1-windowsservercore-ltsc2016/build/Dockerfile
```diff
@@ -3,30 +3,34 @@ FROM microsoft/dotnet-framework:4.7.1-windowsservercore-ltsc2016
...
+ENV NUGET_VERSION 4.4.1
 RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
-    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v4.4.1/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
+    Invoke-WebRequest -UseBasicParsing https://dist.nuget.org/win-x86-commandline/v$Env:NUGET_VERSION/nuget.exe -OutFile $Env:ProgramFiles\NuGet\nuget.exe
 
 # Install VS Test Agent
 RUN Invoke-WebRequest -UseBasicParsing https://download.visualstudio.microsoft.com/download/pr/100232757/8a386d27295953ee79281fd1f1832e2d/vs_TestAgent.exe -OutFile vs_TestAgent.exe; \
     Start-Process vs_TestAgent.exe -ArgumentList '--quiet', '--norestart', '--nocache' -NoNewWindow -Wait; \
     Remove-Item -Force vs_TestAgent.exe; \
 # Install VS Build Tools
```
ravimeda: Nit - a new line before the comment looks cleaner. Same with the other file.

4.7.1-windowsservercore-1709/build/Dockerfile
MichaelSimons: I would rather the build fail in this scenario because something would have changed and I would want to know about it.

4.7.1-windowsservercore-ltsc2016/build/Dockerfile
MichaelSimons: You can't have a new line because the Dockerfile parser doesn't support it.  You would need the line a continuation `\`.  Because of that I am on the fence on whether or not this really adds value.  Leaving as is.

PR #54, opened 2017-12-05
Add manifest.json in order to support using image-builder tools on repo

MichaelSimons:
Also added Get-TagsDocumentation.ps1 script.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,0 +1,140 @@
+{
+  "repos": [
+    {
+      "name": "microsoft/dotnet-framework",
+      "readmePath": "README.dotnet-framework.md",
+      "images": [
+        {
+          "sharedTags": {
+            "4.7.1": {},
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -0,0 +1,140 @@
+{
+  "repos": [
+    {
+      "name": "microsoft/dotnet-framework",
+      "readmePath": "README.dotnet-framework.md",
+      "images": [
+        {
+          "sharedTags": {
+            "4.7.1": {},
```
ravimeda: Clarifying question - Tags will be in descending order and hence there is no need for `readmeOrder` , right?

MichaelSimons: If `readmeOrder` is not specified, then the manifest order is used.  The only scenario in which I would recommend specifying the `readmeOrder` is when the build order is different than your desired order in the readme.

PR #57, opened 2017-12-06
 Switched to use backtick escape character in Dockerfiles

### LINKED ISSUE #46: Consider setting the backtick as the escape character in the Dockerfiles
Using the backtick as the escape character in Windows based Dockerfiles has several advantages.  See the [Dockerfile Reference](https://docs.docker.com/engine/reference/builder/#escape) for details.  While it is not necessary, I believe it would be beneficial for Microsoft Dockerfiles to use the backtick escape character over the default backslash since they are often referenced and copied by many users.  Using the backtick will reduce the chances of users encountering unintentional issues (e.g. `COPY testfile.txt c:\`)

This is already being used in the .NET Core Docker repos.

MichaelSimons:
Fixes #46

### DIFF
--- a/3.5-windowsservercore-1709/build/Dockerfile
+++ b/3.5-windowsservercore-1709/build/Dockerfile
@@ -1,36 +1,38 @@
+# escape=`
+
 FROM microsoft/dotnet-framework:3.5-windowsservercore-1709
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
 # Install NuGet CLI
 ENV NUGET_VERSION 4.4.1
-RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
+RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; `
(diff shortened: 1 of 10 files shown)


3.5-windowsservercore-1709/build/Dockerfile
```diff
@@ -1,36 +1,38 @@
+# escape=`
+
 FROM microsoft/dotnet-framework:3.5-windowsservercore-1709
 
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
 # Install NuGet CLI
 ENV NUGET_VERSION 4.4.1
-RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; \
+RUN New-Item -Type Directory $Env:ProgramFiles\NuGet; `
```
muratg: @natemcmaster @glennc Do we need something for ASP.NET images as well?

MichaelSimons: I believe the ASP.NET Core Windows Dockerfiles are already using the backtick escape - https://github.com/aspnet/aspnet-docker/blob/master/1.1/nanoserver-sac2016/sdk/Dockerfile#L1

natemcmaster: Correct, we already do this.

PR #59, opened 2017-12-06
Refactor 3.5 runtime Dockerfiles

### LINKED ISSUE #45: Update Dockerfiles to use COPY instruction rather than ADD
To conform with [Dockerfile best practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy), the COPY instruction is preferred over ADD.  All Microsoft Dockerfiles should be following the best practices as they are often referred upon and copied.

### LINKED ISSUE #58: 3.5 Runtime images contain unnecessary installer files
The microsoft/dotnet-framework:3.5-windowsservercore-1709 and microsoft/dotnet-framework:3.5-windowsservercore-ltsc2016 images contain installer files that are significant in size.  

```
...
ADD install /build
RUN DISM /Online /Add-Package /PackagePath:C:\build\microsoft-windows-netfx3-ondemand-package.cab & \
    DISM /Online /Add-Package /PackagePath:C:\build\patch\Windows10.0-KB3213986-x64.cab  & \
    rd /s /q C:\build
...
```

Even though the installer files are deleted within the Dockerfile, because they are not deleted within the layer that they were created, the net result is that when the image is pulled the installer files are included within one of the layers.  You can see this by looking at the layers.

```
> docker history microsoft/dotnet-framework:3.5-windowsservercore-ltsc2016
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
0bffe236f9e8        3 weeks ago         cmd /S /C set COMPLUS_NGenProtectedProcess...   985MB
<missing>           3 weeks ago         cmd /S /C DISM /Online /Add-Package /Packa...   1.11GB
<missing>           3 weeks ago         cmd /S /C #(nop) ADD dir:28ab8efa444c3f324...   1.32GB
<missing>           3 weeks ago         Install update 10.0.14393.1884                  2.72GB
<missing>           11 months ago       Apply image 10.0.14393.0                        7.68GB
```

```
> docker history microsoft/dotnet-framework:3.5-windowsservercore-1709
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
cb1fd09e4366        3 weeks ago         cmd /S /C set COMPLUS_NGenProtectedProcess...   1.06GB
<missing>           3 weeks ago         cmd /S /C DISM /Online /Add-Package /Packa...   864MB
<missing>           3 weeks ago         cmd /S /C #(nop) ADD dir:7cb14cb951c64e01e...   464MB
<missing>           4 weeks ago         Install update 10.0.16299.64                    962MB
<missing>           2 months ago        Apply image 10.0.16299.15                       4.62GB
```

MichaelSimons:
These changes address the following:
1.  Remove the unnecessary installer files that are currently included in the 3.5 runtime docker images.  These files are significant in size (464MB for 1709 and 1.32 GB for ltsc2016)
1.  Contents of the 3.5 Dockerfiles are transparent to users.  The LCU being installed in the Dockerfile was not reflective of what was contained in the image.
1.  Dockerfile is now buildable by anyone.  This will allow users to create their own 3.5 image in the case they can't use these 3.5 images.

Fixes #58 
Fixes #45

### DIFF
--- a/3.5-windowsservercore-1709/runtime/Dockerfile
+++ b/3.5-windowsservercore-1709/runtime/Dockerfile
@@ -2,10 +2,34 @@
 
 FROM microsoft/windowsservercore:1709
 
-ADD install /build
-RUN DISM /Online /Add-Package /PackagePath:C:\build\microsoft-windows-netfx3-ondemand-package.cab & `
-    DISM /Online /Add-Package /PackagePath:C:\build\patch\Windows10.0-KB3213986-x64.cab  & `
-    rd /s /q C:\build
+# Install .NET Fx 3.5
+RUN powershell -Command `
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        Invoke-WebRequest `
+            -UseBasicParsing `
+            -Uri "https://dotnetbinaries.blob.core.windows.net/dockerassets/microsoft-windows-netfx3-1709.zip" `
+            -OutFile microsoft-windows-netfx3.zip; `
+        Expand-Archive microsoft-windows-netfx3.zip; `
+        Remove-Item -Force microsoft-windows-netfx3.zip; `
+        Add-WindowsPackage -Online -PackagePath .\microsoft-windows-netfx3\microsoft-windows-netfx3-ondemand-package.cab; `
+        Remove-Item -Force -Recurse microsoft-windows-netfx3
+
+# Apply latest patch
+RUN powershell -Command `
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        Invoke-WebRequest `
+            -UseBasicParsing `
+            -Uri "http://download.windowsupdate.com/d/msdownload/update/software/secu/2017/11/windows10.0-kb4048955-x64_f214597fd57cd8284a363ada825a3d8ca16bf632.msu" `
+            -OutFile patch.msu; `
+        New-Item -Type Directory patch; `
+        Start-Process expand -ArgumentList 'patch.msu', 'patch', '-F:*' -NoNewWindow -Wait; `
+        Remove-Item -Force patch.msu; `
+        Add-WindowsPackage -Online -PackagePath C:\patch\Windows10.0-KB4048955-x64.cab; `
+        Remove-Item -Force -Recurse \patch
+
+# ngen .NET Fx
 RUN set COMPLUS_NGenProtectedProcess_FeatureEnabled=0 & `
     \Windows\Microsoft.NET\Framework64\v4.0.30319\ngen update & `
     \Windows\Microsoft.NET\Framework\v4.0.30319\ngen update & `
--- a/3.5-windowsservercore-1709/runtime/Dockerfile
+++ b/3.5-windowsservercore-1709/runtime/Dockerfile
@@ -21,12 +21,12 @@ RUN powershell -Command `
         $ProgressPreference = 'SilentlyContinue'; `
         Invoke-WebRequest `
             -UseBasicParsing `
-            -Uri "http://download.windowsupdate.com/d/msdownload/update/software/secu/2017/11/windows10.0-kb4048955-x64_f214597fd57cd8284a363ada825a3d8ca16bf632.msu" `
+            -Uri "http://download.windowsupdate.com/d/msdownload/update/software/secu/2017/12/windows10.0-kb4054517-x64_f1c0e8eae96b357bc13e5668c5874904502106df.msu" `
             -OutFile patch.msu; `
         New-Item -Type Directory patch; `
         Start-Process expand -ArgumentList 'patch.msu', 'patch', '-F:*' -NoNewWindow -Wait; `
         Remove-Item -Force patch.msu; `
-        Add-WindowsPackage -Online -PackagePath C:\patch\Windows10.0-KB4048955-x64.cab; `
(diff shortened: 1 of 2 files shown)


3.5-windowsservercore-1709/runtime/Dockerfile
```diff
@@ -1,11 +1,37 @@
...
+            -UseBasicParsing `
+            -Uri "https://dotnetbinaries.blob.core.windows.net/dockerassets/microsoft-windows-netfx3-1709.zip" `
+            -OutFile microsoft-windows-netfx3.zip; `
+        Expand-Archive microsoft-windows-netfx3.zip; `
+        Remove-Item -Force microsoft-windows-netfx3.zip; `
+        Add-WindowsPackage -Online -PackagePath .\microsoft-windows-netfx3\microsoft-windows-netfx3-ondemand-package.cab; `
+        Remove-Item -Force -Recurse microsoft-windows-netfx3
+
+# Apply latest patch
+RUN powershell -Command `
```
shirhatti: Why download the file in the `RUN` directive as opposed to directly using the `ADD` directive?

3.5-windowsservercore-1709/runtime/Dockerfile
```diff
@@ -1,11 +1,37 @@
...
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        Invoke-WebRequest `
+            -UseBasicParsing `
+            -Uri "http://download.windowsupdate.com/d/msdownload/update/software/secu/2017/11/windows10.0-kb4048955-x64_f214597fd57cd8284a363ada825a3d8ca16bf632.msu" `
```
shirhatti: The Uri should be a parameter. When building candidate images we point to a local file share instead

3.5-windowsservercore-1709/runtime/Dockerfile
```diff
@@ -1,11 +1,37 @@
...
+RUN powershell -Command `
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        Invoke-WebRequest `
+            -UseBasicParsing `
+            -Uri "https://dotnetbinaries.blob.core.windows.net/dockerassets/microsoft-windows-netfx3-1709.zip" `
+            -OutFile microsoft-windows-netfx3.zip; `
+        Expand-Archive microsoft-windows-netfx3.zip; `
+        Remove-Item -Force microsoft-windows-netfx3.zip; `
+        Add-WindowsPackage -Online -PackagePath .\microsoft-windows-netfx3\microsoft-windows-netfx3-ondemand-package.cab; `
```
shirhatti: Add the `-NoRestart` flag

3.5-windowsservercore-1709/runtime/Dockerfile
```diff
@@ -1,11 +1,37 @@
...
+        $ErrorActionPreference = 'Stop'; `
+        $ProgressPreference = 'SilentlyContinue'; `
+        Invoke-WebRequest `
+            -UseBasicParsing `
+            -Uri "http://download.windowsupdate.com/d/msdownload/update/software/secu/2017/11/windows10.0-kb4048955-x64_f214597fd57cd8284a363ada825a3d8ca16bf632.msu" `
+            -OutFile patch.msu; `
+        New-Item -Type Directory patch; `
+        Start-Process expand -ArgumentList 'patch.msu', 'patch', '-F:*' -NoNewWindow -Wait; `
+        Remove-Item -Force patch.msu; `
+        Add-WindowsPackage -Online -PackagePath C:\patch\Windows10.0-KB4048955-x64.cab; `
```
shirhatti: Add the `-NoRestart` flag

3.5-windowsservercore-1709/runtime/Dockerfile
MichaelSimons: See https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy
>Because image size matters, using ADD to fetch packages from remote URLs is strongly discouraged; you should use curl or wget instead. That way you can delete the files you no longer need after they’ve been extracted and you won’t have to add another layer in your image.

3.5-windowsservercore-1709/runtime/Dockerfile
MichaelSimons: I’m not certain we should be specify the /NoRestart flag.  Whatever we install should be finalized.  You don’t want any installation activities to happen each time the container is started (I’m not certain this is even possible).  By not specifying /NoRestart, the Docker build will hang/fail if/when a restart is required, thus we will know something is "wrong".  When /NoRestart is specified, the build completes and a warning is emitted stating there is a pending restart.

3.5-windowsservercore-1709/runtime/Dockerfile
MichaelSimons: I had thought about this but how would you get the credentials to access the file share within the container?  The build container is not domain joined and is running as the ContainerAdministrator.

shirhatti: Off the top of my head, I don't if docker build args support default values, but I think that would be the best case for us.

MichaelSimons: Dockerfile ARGS support default values.  The part I am missing is that is you pass in a local file share url at build time, how is the build container going to get access to it?

shirhatti: I don't have a good answer 😕
@mcy94w Thoughts?

richlander:
LGTM

PR #65, opened 2018-01-04
Build and test images

### LINKED ISSUE #66: Enable CI
CI should be enabled on this repo to ensure Dockerfiles are always in a buildable state are changes are made.

ravimeda:
Unit tests to validate the framework images are needed.  Such validation is available in [dotnet-docker](https://github.com/dotnet/dotnet-docker).  This PR lays the foundation to build and test the framework images. Current changes use [ImageBuilder](https://github.com/dotnet/docker-tools/tree/master/Microsoft.DotNet.ImageBuilder) to build the framework images locally and in CI.  A subsequent PR will bring the unit tests.

  Fixes #66

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -1,34 +1,20 @@
 [cmdletbinding()]
 param(
-    [string]$VersionFilter,
-    [string]$OSFilter,
+    [string]$VersionFilter = "*",
+    [string]$OSFilter = "*",
     [switch]$IsDryRun
 )
 
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
 
-$buildFilter = "*"
-if (-not [string]::IsNullOrEmpty($VersionFilter))
-{
-    $buildFilter = "$VersionFilter-$buildFilter"
-}
-if (-not [string]::IsNullOrEmpty($OSFilter))
-{
-    if ([string]::IsNullOrEmpty($VersionFilter))
-    {
-        $buildFilter = "$buildFilter-$OSFilter/*"
-    }
-    else
-    {
-        $buildFilter = "$buildFilter$OSFilter/*"
-    }
-}
+$buildFilter = "$VersionFilter-$OSFilter/*"
 
 function Get-ImageBuilder()
 {
-    $imageBuilderImageName = 'microsoft/dotnet-buildtools-prereqs:image-builder-nanoserver-20180103080524'
+    $imageBuilderImageName = 'microsoft/dotnet-buildtools-prereqs:image-builder-nanoserver-20180105103709'
     $imageBuilderContainerName = "ImageBuilder-$(Get-Date -Format yyyyMMddhhmmss)"
+    New-Item -Path "$imageBuilderFolder" -ItemType Directory -Force | Out-Null
 
     $attempt = 0
     $maxRetries = 5
@@ -36,12 +22,7 @@ function Get-ImageBuilder()
     while ($attempt -lt $maxRetries)
     {
         try {
-            New-Item -Path "$imageBuilderFolder" -ItemType Directory -Force | Out-Null
-
             Invoke-Expression "docker pull $imageBuilderImageName"
-            Invoke-Expression "docker create --name $imageBuilderContainerName $imageBuilderImageName"
-            Invoke-Expression "docker cp ${imageBuilderContainerName}:/image-builder $imageBuilderFolder"
-            Invoke-Expression "docker rmi -f $imageBuilderImageName"
             break
         }
         catch {
@@ -60,16 +41,21 @@ function Get-ImageBuilder()
             throw "Retry ${attempt}/${maxRetries} failed, no more retries left."
         }
     }
+
+    Invoke-Expression "docker create --name $imageBuilderContainerName $imageBuilderImageName"
+    # Copying the 'image-builder' folder and not just the content due to https://github.com/moby/moby/issues/34638
+    Invoke-Expression "docker cp ${imageBuilderContainerName}:/image-builder $imageBuilderFolder"
+    Invoke-Expression "docker rmi -f $imageBuilderImageName"
 }
 
 $imageBuilderFolder = [System.IO.Path]::Combine("$PSScriptRoot", ".Microsoft.DotNet.ImageBuilder")
-$imageBuilderExe = [System.IO.Path]::Combine($imageBuilderFolder, "image-builder", "Microsoft.DotNet.ImageBuilder.exe")
+$imageBuilderExe = [System.IO.Path]::Combine("$imageBuilderFolder", "image-builder", "Microsoft.DotNet.ImageBuilder.exe")
 if (-not (Test-Path -Path "$imageBuilderExe" -PathType Leaf))
 {
     Get-ImageBuilder
 }
 
-$imageBuilderArgs = 'build --path "$buildFilter"'
+$imageBuilderArgs = "build --path $buildFilter "
 if ($IsDryRun)
 {
     $imageBuilderArgs += " --dry-run"
@@ -79,3 +65,5 @@ Invoke-Expression "$imageBuilderExe $imageBuilderArgs"
 if ($LastExitCode -ne 0) {
     throw "Failed executing $imageBuilderExe $imageBuilderArgs"
 }
+
+./tests/run-tests.ps1 -VersionFilter $VersionFilter -OSFilter $OSFilter
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -8,8 +8,6 @@ param(
 Set-StrictMode -Version Latest
 $ErrorActionPreference = 'Stop'
 
-$(docker version) | ForEach-Object { Write-Host "$_" }
-
 $buildFilter = "*"
 if (-not [string]::IsNullOrEmpty($VersionFilter))
 {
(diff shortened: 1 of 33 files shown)


build-and-test.ps1
```diff
@@ -0,0 +1,46 @@
...
+param(
+    [string]$VersionFilter,
+    [string]$OSFilter,
+    [switch]$IsDryRun
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$(docker version) | ForEach-Object { Write-Host "$_" }
```
ravimeda: Docker version is emitted here and in ImageBuilder. Either move this statement into the `if` block where ImageBuilder image is pulled (or) remove version emit in ImageBuilder itself.  Any preferences?

build-and-test.ps1
```diff
@@ -0,0 +1,46 @@
...
+
+$buildFilter = "*"
+if (-not [string]::IsNullOrEmpty($VersionFilter))
+{
+    $buildFilter = "$VersionFilter-$buildFilter"
+}
+if (-not [string]::IsNullOrEmpty($OSFilter))
+{
+    $buildFilter = "$buildFilter$OSFilter/*"
```
ravimeda: `-OSFilter 1709` would match all 1709 Dockerfiles.  So one might expect `-OSFilter ltsc` would match all `ltsc`. But it wouldn't.  This is due to `2016` suffix, which means `-OSFilter ltsc2016` is how the parameter value should be for LTSC images.
Any preferences on the usage?

netci.groovy
```diff
@@ -0,0 +1,32 @@
+import jobs.generation.Utilities
+
+def project = GithubProject
+def branch = GithubBranchName
+def isPR = true
+def platformList = ['Windows_2016:WindowsServerCore-ltsc2016', 'Windows_2016:WindowsServerCore-1709']
+def versionList = ['3.5', '4.6.2', '4.7.1']
```
ravimeda: Each time a new version is added/removed, one has to edit this list. In a way it is good to revisit the coverage but there is maintenance cost.

build-and-test.ps1
MichaelSimons: Seems like there should be a `-` if an OSFilter is specified but not a VersionFilter - e.g. $buildFilter-$OSFilter/*.

I think it is just fine that `-OSFilter ltsc` is not supported.  The user can specify `*ltsc*` for this to work.

netci.groovy
MichaelSimons: Just realized there is an issue here - 3.5 and 4.7.1 are the only versions supported on 1709.  You can see it in https://github.com/Microsoft/dotnet-framework-docker/blob/master/build-pipeline/pipeline.json.  CI is going to have to account for this.

build-and-test.ps1
ravimeda: Agree. My thinking earlier was - Docker is a prerequisite for the build-test experience.  If a prerequisite is not available, in this case Docker version command fails, then exit immediately.

build-and-test.ps1
ravimeda: `*ltsc*` sounds fine.

Hyphen `-` is not required. But would make it more accurate. I will update this as follows -

```
$buildFilter = "*"
if (-not [string]::IsNullOrEmpty($VersionFilter))
{
    $buildFilter = "$VersionFilter-$buildFilter"
}
if (-not [string]::IsNullOrEmpty($OSFilter))
{
    if ([string]::IsNullOrEmpty($VersionFilter))
    {
        $buildFilter = "$buildFilter-$OSFilter/*"
    }
    else
    {
        $buildFilter = "$buildFilter$OSFilter/*"
    }
}
```

build-and-test.ps1
```diff
@@ -0,0 +1,46 @@
...
+}
+
+$imageBuilderFolder = ".Microsoft.DotNet.ImageBuilder"
```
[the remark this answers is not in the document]
ravimeda: Ok. Will update.

build-and-test.ps1
```diff
@@ -0,0 +1,46 @@
...
+if (-not (Test-Path -Path "$imageBuilderFolder" -PathType Container))
```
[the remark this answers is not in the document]
ravimeda: Good point! Will update accordingly.

build-and-test.ps1
```diff
@@ -0,0 +1,46 @@
...
+    $imageBuilderContainerName = "ImageBuilder-$(Get-Date -Format yyyyMMddhhmmss)"
+
+    Invoke-Expression "docker pull $imageBuilderImageName"
+    Invoke-Expression "docker create --name $imageBuilderContainerName $imageBuilderImageName"
+    Invoke-Expression "docker cp ${imageBuilderContainerName}:/image-builder $imageBuilderFolder"
+    Invoke-Expression "docker rmi -f $imageBuilderImageName"
+}
+
+$imageBuilderExe = [System.IO.Path]::Combine($imageBuilderFolder, "image-builder", "Microsoft.DotNet.ImageBuilder.exe")
+$imageBuilderArgs = 'build --repo microsoft/dotnet-framework --path "$buildFilter"'
```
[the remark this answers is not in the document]
ravimeda: You are correct. `repo` is not needed for build command. It is needed for other commands such as the one that generates and updates readme.

netci.groovy
```diff
@@ -0,0 +1,32 @@
+import jobs.generation.Utilities
```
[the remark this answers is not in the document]
ravimeda: Done.

netci.groovy
```diff
@@ -0,0 +1,32 @@
...
+def isPR = true
+def platformList = ['Windows_2016:WindowsServerCore-ltsc2016', 'Windows_2016:WindowsServerCore-1709']
+def versionList = ['3.5', '4.6.2', '4.7.1']
+
+platformList.each { platform ->
+    def(hostOS, containerOS) = platform.tokenize(':')
+    def machineLabel = 'latest-docker'
+
+    versionList.each { version ->
+        def newJobName = Utilities.getFullJobName(project, "${containerOS}_${version}", isPR)
```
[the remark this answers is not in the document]
ravimeda: Ok. will update.

netci.groovy
ravimeda: Good point! I will update accordingly.

build-and-test.ps1
```diff
@@ -0,0 +1,81 @@
...
+
+    $attempt = 0
+    $maxRetries = 5
+    $waitFactor = 6
+    while ($attempt -lt $maxRetries)
+    {
+        try {
+            New-Item -Path "$imageBuilderFolder" -ItemType Directory -Force | Out-Null
+
+            Invoke-Expression "docker pull $imageBuilderImageName"
```
ravimeda: As discussed offline, it would be simpler if ImageBuilder binaries were available as a compressed file hosted in Azure storage.  Acquiring ImageBuilder would then be simpler - download and extract,

ravimeda: Logged https://github.com/dotnet/docker-tools/issues/67

netci.groovy
```diff
@@ -0,0 +1,38 @@
...
+def project = GithubProject
+def branch = GithubBranchName
+def isPR = true
+def platformList = ['Windows_2016:WindowsServerCore-ltsc2016', 'Windows_2016:WindowsServerCore-1709']
+
+platformList.each { platform ->
+    def(hostOS, containerOS) = platform.tokenize(':')
+    def machineLabel = 'latest-docker'
+
+    if (containerOS == 'WindowsServerCore-1709') {
```
MichaelSimons: Per https://github.com/Microsoft/dotnet-framework-docker/commit/1f612825330fd27a99d4f20c6a68745b274e6bac, this can now be updated to a single static list - ['3.5', '4.*']

build-and-test.ps1
```diff
@@ -0,0 +1,81 @@
...
+            Remove-Item -Path "$imageBuilderFolder" -Recurse -Force -ErrorAction Continue
+        }
+        else {
+            throw "Retry ${attempt}/${maxRetries} failed, no more retries left."
+        }
+    }
+}
+
+$imageBuilderFolder = [System.IO.Path]::Combine("$PSScriptRoot", ".Microsoft.DotNet.ImageBuilder")
+$imageBuilderExe = [System.IO.Path]::Combine($imageBuilderFolder, "image-builder", "Microsoft.DotNet.ImageBuilder.exe")
```
MichaelSimons: Why create the image-builder sub folder?  This seems to add unnecessary nesting.

build-and-test.ps1
MichaelSimons: I think the `docker pull` is the only thing that should be in the retry.  It is the only call that is intermittent.  If the other cmds fail, something is drastically wrong and a retry isn't going to succeed..

build-and-test.ps1
```diff
@@ -0,0 +1,81 @@
...
+if (-not (Test-Path -Path "$imageBuilderExe" -PathType Leaf))
+{
+    Get-ImageBuilder
+}
+
+$imageBuilderArgs = 'build --path "$buildFilter"'
```
MichaelSimons: Now that the tests have been checked in, you will need to pass the OSFilter and VersionFilter as build-args.

build-and-test.ps1
```diff
@@ -0,0 +1,81 @@
...
+    [string]$VersionFilter,
+    [string]$OSFilter,
+    [switch]$IsDryRun
+)
+
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+$buildFilter = "*"
+if (-not [string]::IsNullOrEmpty($VersionFilter))
```
MichaelSimons: What do you think about defaulting VersionFilter and OSFilter to '*' if not specified?  Then this logic could be simplified to - `buildFilter = "{VersionFilter}-{OSFilter}/*"`

netci.groovy
ravimeda: Ok. Updated as `def versionList = ['3.5', '4.*']`

build-and-test.ps1
ravimeda: Yes, this works. Thanks!

build-and-test.ps1
ravimeda: Ok. Will move rest of the command outside wait-retry block.

build-and-test.ps1
ravimeda: As discussed offline, this is due to https://github.com/moby/moby/issues/34638

build-and-test.ps1
ravimeda: Clarification - did you mean invoke tests as in -
`./tests/run-test.ps1 -VersionFilter $VersionFilter -OSFilter $OSFilter`

MichaelSimons: I mean you will need to pass in `--test-var VersionFilter=$VersionFilter --test-var OSFilter=$OSFilter' into ImageBuilder

build-and-test.ps1
```diff
@@ -0,0 +1,69 @@
...
+
+$buildFilter = "$VersionFilter-$OSFilter/*"
+
+function Get-ImageBuilder()
+{
+    $imageBuilderImageName = 'microsoft/dotnet-buildtools-prereqs:image-builder-nanoserver-20180105103709'
+    $imageBuilderContainerName = "ImageBuilder-$(Get-Date -Format yyyyMMddhhmmss)"
+    New-Item -Path "$imageBuilderFolder" -ItemType Directory -Force | Out-Null
+
+    $attempt = 0
```
MichaelSimons: I think it would help the readability to factor out the logic to perform the retry into a separate function.

build-and-test.ps1
ravimeda: Ok. I didn't notice the recent commit (e618fa116d493746b908b557cfbd7d6ea583c8c2) that introduced `testCommands` section in the manifest.  Yes, I will pass those args.

build-and-test.ps1
```diff
@@ -10,28 +10,26 @@ $ErrorActionPreference = 'Stop'
 
 $buildFilter = "$VersionFilter-$OSFilter/*"
 
-function Get-ImageBuilder()
-{
-    $imageBuilderImageName = 'microsoft/dotnet-buildtools-prereqs:image-builder-nanoserver-20180105103709'
-    $imageBuilderContainerName = "ImageBuilder-$(Get-Date -Format yyyyMMddhhmmss)"
-    New-Item -Path "$imageBuilderFolder" -ItemType Directory -Force | Out-Null
+function ExecuteCommand {
```
MichaelSimons: I think it would be helpful if the name called out that that retry logic exists.  This helps clarify why the function is used over Invoke-Expression.  e.g. ExecuteWithRetry

build-and-test.ps1
```diff
@@ -50,20 +54,16 @@ function Get-ImageBuilder()
...
-{
+if (-not (Test-Path -Path "$imageBuilderExe" -PathType Leaf)) {
     Get-ImageBuilder
 }
 
-$imageBuilderArgs = "build --path $buildFilter "
-if ($IsDryRun)
-{
+$imageBuilderArgs = "build --path $buildFilter --test-var VersionFilter=$VersionFilter --test-var OSFilter=$OSFilter"
+if ($IsDryRun) {
```
MichaelSimons: What do you think about replacing the IsDryRun parameter with a string CustomImageBuilderArgs?  This would make this script more robust so that it could be used within the official builds.  The CustomImageBuilderArgs could be used to specify the `push` related arguments.

build-and-test.ps1
ravimeda: I did consider the name `ExecuteWithRetry`. However, with the word `Retry` the questions that come to my (user's) mind is - how do I set the number of retries, wait time? Is the wait time in seconds or minutes, etc.

build-and-test.ps1
ravimeda: To clarify - change the type of the parameter to `string`, and append the parameter value to the default args.  So in that case, I think `ImageBuilderCustomArgs` would convey better, and the logic would be -

```
$imageBuilderArgs = "build --path $buildFilter --test-var VersionFilter=$VersionFilter --test-var OSFilter=$OSFilter"
if ([string]::IsNullOrWhiteSpace($ImageBuilderCustomArgs)) {
    $imageBuilderArgs += " $ImageBuilderCustomArgs"
}
```

MichaelSimons: yes that is what I was thinking - and I like your suggested name.

build-and-test.ps1
MichaelSimons: Since this is a local method, I think it is reasonable for the reader to look at the implementation.  I think `ExecuteWithRetry` addresses the point I raised on why the method exists over using the built in Invoke-Expression and why this method is only being used for the pull cmd.

ravimeda:
@dotnet-bot test cli please

ravimeda:
@dotnet-bot test ci please

ravimeda:
@dotnet-bot test ci please

ravimeda:
@dotnet-bot test ci please

ravimeda:
Thanks @mmitche for enabling CI!

PR #76, opened 2018-01-12
Remove hard coded repo owner from tests

MichaelSimons:
This allows the repo to easily get built and pushed to an Azure Container Registry which is needed at times for testing purposes.


tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -18,6 +18,7 @@ public class ImageTests
         private const string WSC_1709 = "windowsservercore-1709";
 
         private static string OSFilter => Environment.GetEnvironmentVariable("IMAGE_OS_FILTER");
+        private static string RepoOwner => Environment.GetEnvironmentVariable("REPO_OWNER") ?? "Microsoft";
```
ravimeda: Lower case 'm' as in `microsoft`?

MichaelSimons: Good catch - fixed

PR #120, opened 2018-04-18
adding web testing projects

HongGit:
Add Web testing projects.


HongGit:
Done.

@MichaelSimons this PR contains all test projects.   I have verified each DockerFile and made sure that they work properly.

MichaelSimons:
@HongGit - Can you summarize the next steps?  I am assuming it is to implement some automation.

HongGit:
@MichaelSimons and @richlander I am done integrating web testing projects into the main project.  

Feel free to let me know if anything is missing.

HongGit:
@MichaelSimons it seems that 4.6.2 Build directory is missing in the build branch, is that by design?  

In addition, please let me know once you have 4.7.2 added, I will add the corrsponding web projects as well.

MichaelSimons:
@HongGit - I updated the aspnet-build branch with the latest changes from master.  There is no 4.6.2 or 4.7 sdk images.  4.7.2 is the preferred sdk to use to build these projects.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -93,6 +93,39 @@ public void VerifyImages(ImageDescriptor imageDescriptor)
             }
         }
 
+        [Theory]
+        [MemberData(nameof(GetVerifyImagesData))]
+        public void VerifyImagesWithWebApps(ImageDescriptor imageDescriptor)
```
MichaelSimons: Seems to be a lot of duplication with the other test case.  Consider refactoring to encapsulate the commonality in a single location.

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-3.5/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BASE_BUILD_IMAGE
+
+FROM $BASE_BUILD_IMAGE
+
+WORKDIR /app
+COPY . ./
+
+RUN msbuild SimpleWebApplication.csproj /verbosity:quiet /p:Configuration=Release
+
+ENTRYPOINT dir ./bin/SimpleWebApplication.dll
```
MichaelSimons: The ENTRYPOINT doesn't seem like a real usage scenario.  If you want to capture this, it seems like it would be better done with a `docker run`.

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-3.5/Dockerfile
```diff
@@ -0,0 +1,10 @@
+ARG BASE_BUILD_IMAGE
+
+FROM $BASE_BUILD_IMAGE
+
+WORKDIR /app
+COPY . ./
+
+RUN msbuild SimpleWebApplication.csproj /verbosity:quiet /p:Configuration=Release
```
MichaelSimons: Does the `/verbosity:quiet` make it difficult to diagnose build failures?

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-3.5/Dockerfile
MichaelSimons: I think the more canonical way to do this would be do use `docker run` after you built your Dockerfile - `docker run dir ./bin/SimpleWebApplication.dll`

MichaelSimons: >I am trying to make sure that that build was successful.

How is this being verified?  I'm not seeing any automation that verifies the build was successful other than the Dockerfile built without error.  Perhaps that is adequate, I am trying to understand.

MichaelSimons: > I don't see how "docker run" is different from using CMD or ENTRYPOINT. I reverted to my original approach, which is using dir, which verifies right after msbuild.

The responsibility of the Dockerfile in a scenario like this is to build the application.  Setting the ENTRYPOINT to list the contents on a directory doesn't make sense and makes the resulting image more difficult to use for anything else as you would need to override the ENTRYPOINT. 

Where is the new PR?  I am not finding it?  Why not just update this PR?

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-4.6.2/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+FROM $BASE_BUILD_IMAGE
+
+WORKDIR /app
+COPY . ./
+
+RUN nuget restore SimpleWebApplication.csproj -SolutionDirectory -NoCache
+
+RUN msbuild SimpleWebApplication.csproj /p:Configuration=Release
+
+RUN dir ./bin/SimpleWebApplication.dll
```
MichaelSimons: Per our discussion, I think the verification (e.g. `dir ./bin/SimpleWebApplication.dll`) logic best belongs in the test case.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -59,40 +59,66 @@ private static string GetFilterRegexPattern(string filter)
...
+                 {
+                        $"BASE_BUILD_IMAGE={baseBuildImage}",
+                        $"BASE_RUNTIME_IMAGE={baseRuntimeImage}",
+                 };
+            }
+            else
+            {
+                appBuildArgs = new string[]
+                 {
+                        $"BASE_BUILD_IMAGE={baseBuildImage}",
```
MichaelSimons: I think it would be best to not have this duplicated.  You could change appBuildArgs to a list and initialize it with the build image.  The includeRuntime case could then add the runtime image.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -59,40 +59,66 @@ private static string GetFilterRegexPattern(string filter)
...
-                    });
+                    buildArgs: appBuildArgs);
 
                 DockerHelper.Run(image: appId, containerName: appId);
             }
             finally
             {
                 DockerHelper.DeleteImage(appId);
             }
+
```
MichaelSimons: There are a couple WS changes?  Were those intentional?

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-4.6.2/Dockerfile
HongGit: I gave this more thoughts.  CMD is probably the most appropriate, which is executed during "docker run", that will be the test case you are referring.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
HongGit: I fixed a merge conflict, is this still an issue?

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-4.6.2/Dockerfile
MichaelSimons: I don't understand why it would be most appropriate for the CMD to be `dir ./bin/SimpleWebApplication.dll`.  Please explain.  This doesn't seem like a normal usage.  I would prefer the Dockerfile to be a typical usage we expect to see.  All verification logic should reside within the tests.  Please fix so that we can merge this PR.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
MichaelSimons: yes it is.  You can see it very easily in the diff.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
MichaelSimons: ping

HongGit: This will be addressed in the next PR update.

HongGit: What is WS?  Sorry, I don't see it.  All the changes in ImageTests.cs are intentional.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
HongGit: What is WS? Sorry, I don't see it. All the changes in ImageTests.cs are intentional.

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/webapp-4.6.2/Dockerfile
HongGit: using CMD, the existence of the dll after build is being verified at docker run time.  The verification logic is in the testcase, reside within the tests.  Hopefully this answers your questions!

MichaelSimons: Why is using the CMD most appropriate over using `docker run` on the resulting image.  Per my original comments this is not going to be typical usage.  Please change to place the verification logic in the test case.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
MichaelSimons: WS is whitespace.  Your changes include new blank lines that I don't think add value and make the styling inconsistent.

HongGit: Got it, it will be removed.

tests/Microsoft.DotNet.Framework.Docker.Tests/DockerHelper.cs
```diff
@@ -17,12 +18,12 @@ public DockerHelper(ITestOutputHelper outputHelper)
             OutputHelper = outputHelper;
         }
 
-        public void Build(string tag, string dockerfile, string buildContextPath, params string[] buildArgs)
+        public void Build(string tag, string dockerfile, string buildContextPath, StringCollection buildArgs)
```
MichaelSimons: IEnumerable<string> seems more appropriate as there no dependency on StringCollection.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -70,38 +71,27 @@ private static string GetFilterRegexPattern(string filter)
...
-            VerifyImages(imageDescriptor, "dotnetapp", true);
+            VerifyImages(imageDescriptor, "dotnetapp", "", true);
         }
 
         [Theory]
         [MemberData(nameof(GetVerifyImagesData))]
         public void VerifyImagesWithWebApps(ImageDescriptor imageDescriptor)
         {
-            VerifyImages(imageDescriptor, "webapp", false);
+            VerifyImages(imageDescriptor, "webapp", $"powershell -command \"dir ./bin/SimpleWebApplication.dll\"", false);
```
MichaelSimons: * String interpolation is not being used therefore the `$` is not necessary.

HongGit:
@MichaelSimons  please review the latest update.  Will you please merge and approve this PR so we could get this done?

tests/Microsoft.DotNet.Framework.Docker.Tests/DockerHelper.cs
```diff
@@ -18,7 +18,7 @@ public DockerHelper(ITestOutputHelper outputHelper)
             OutputHelper = outputHelper;
         }
 
-        public void Build(string tag, string dockerfile, string buildContextPath, StringCollection buildArgs)
+        public void Build(string tag, string dockerfile, string buildContextPath, List<string> buildArgs)
```
MichaelSimons: Why the need to take a dependency on List  This method just needs to enumerate?  IEnumerable{string} allows the caller to pass in a variety of types - arrays, lists, collection, etc..

HongGit: There was not an easy way to add to IEnumerable <string>, that is why I used List here.

MichaelSimons: What is wrong with 

```
public void Build(string tag, string dockerfile, string buildContextPath, IEnumerable<string> buildArgs)
```

HongGit: There is nothing wrong with that.  What I need here is to initialize this collection with one string, and in some situation, Add another string to it.  I did not see that IEnumerable supports Add method.

MichaelSimons: `List<string>` implements `IEnumerable<string>` so you will be able to pass in a `List<string>` type into a method that has an `IEnumerable<string>` parameter.

tests/Microsoft.DotNet.Framework.Docker.Tests/DockerHelper.cs
```diff
@@ -17,12 +18,12 @@ public DockerHelper(ITestOutputHelper outputHelper)
...
         }
 
-        public void Build(string tag, string dockerfile, string buildContextPath, params string[] buildArgs)
+        public void Build(string tag, string dockerfile, string buildContextPath, IEnumerable<string> buildArgs)
         {
             string buildArgsOption = null;
             if (buildArgs != null)
             {
-                foreach (string arg in buildArgs)
+                foreach (Object arg in buildArgs)
```
MichaelSimons: This change should be reverted with your change to use `IEnumerable<string>`

PR #121, opened 2018-04-25
Added .NET 4.7.1 SDK

jmezach:
At the moment the build images does not contain the appropriate .NET Framework SDK, which means that those tools (as listed [here](https://docs.microsoft.com/en-us/dotnet/framework/tools/)) aren't available. Since this is a build image it made sense to me to include it.

I've only modified the 4.7.1-windowsservercore-ltsc2016 image so far, but I'm more than happy to modify the others as well if necessary. Wasn't entirely sure on the workflow here.

I also had to put in a workaround for the Visual Studio Installer trying to ngen 32-bit assemblies using the 64-bit version of ngen, which obviously doesn't work.


MichaelSimons:
@jmezach, Can you tell how much this adds to the image size?

jmezach:
Of course. The original microsoft/dotnet-framework-build tagged 4.7.1-windowsservercore-ltsc2016 is 14.3 GB on my box. With the SDK included this goes to 14.6 GB, so it adds about 300 meg.

VishalMadhvani:
Any updates on this?

dcormier:
They probably won't look at it again until conflicts are resolved. @jmezach?

PR #122, opened 2018-04-26
[WIP] Add samples

richlander:
Samples exist in https://github.com/microsoft/dotnet-framework-docker-samples/ today. We're going to close that repo, just like we did with https://github.com/dotnet/dotnet-docker-samples/.

- [x] .NET Framework console sample (dotnetapp)
- [x] ASP.NET Web Forms sample (aspnetapp)
- [x] ASP.NET MVC sample (aspnetmvcapp)
- [x] Samples README
- [x] Samples README for DockerHub

PR #126, opened 2018-05-01
Add samples to this repo

richlander:
Replacing / improving samples at https://github.com/microsoft/dotnet-framework-docker-samples/

Replacing #122


samples/dotnetapp/dotnet-docker-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+
+```console
+docker build -t dotnetapp .
+```
+
+### Run unit tests as part of `docker run`
+
+You can run [unit tests](tests) as part of `docker run` using the following commands. Running tests in this way is useful to get complete tests results for Docker images. The [sample Dockerfile](Dockerfile) exposes multiple [Dockerfile stages](https://docs.docker.com/engine/reference/commandline/build/#specifying-target-build-stage-target) that you can separately target as part of `docker build` and run. The sample Dockerfile includes a `testrunner` stage with a separate `ENTRYPOINT` for unit testing, which is required to maintain a single Dockerfile.
+
+The following commands rely on [volume mounting](https://docs.docker.com/engine/admin/volumes/volumes/) (that's the `-v` argument in the following commands) to enable the test runner to write test log files to your local drive. Without that, running tests as part of `docker run` isn't as useful. You may need to [Enable shared drives](https://docs.docker.com/docker-for-windows/#shared-drives) first.
```
MichaelSimons: The comment on shared drives isn't applicable since these are Windows containers.  I am removing it.

samples/dotnetapp/utils/utils.csproj
```diff
@@ -0,0 +1,7 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
MichaelSimons: Is this intentionally different than the other two projects?

richlander: The other two need to be netcoreapp. This one has the potenial to be netstandard. It aligns with our netstandard-first guidance for libraries.

Issue #129, opened 2018-05-02
Migrate WCF docker samples from dotnet-framework-docker-samples repro to this repro


Lxiamail:
Let's follow the same format as these [samples](https://github.com/Microsoft/dotnet-framework-docker/tree/master/samples).

yujayee:
@MichaelSimons @richlander @Lxiamail  Are there any specific requirements or new format we need to follow when migrating WCF samples?

yujayee:
Close this issue since the pr is merged.

PR #136, opened 2018-05-08
Add WCF sample

yujayee:
#129 
I just copied the WCF sample from old repo to here. Please let me know if there are any new formats we need to follow. Thanks!
@MichaelSimons @richlander @Lxiamail

### DIFF
--- a/samples/wcfapp/Dockerfile.client
+++ b/samples/wcfapp/Dockerfile.client
@@ -1,8 +1,9 @@
 FROM microsoft/dotnet-framework:4.7.2-sdk AS build
 WORKDIR /app
-COPY WcfClient/. ./WcfClient/
-WORKDIR /app/WcfClient
-RUN msbuild
+COPY WcfClient/. .
+RUN msbuild /p:Configuration=Release
 
-FROM microsoft/dotnet-framework:4.7.2-runtime
-COPY --from=build /app/WcfClient/bin/Debug/ ./
+FROM microsoft/dotnet-framework:4.7.2-runtime AS runtime
+WORKDIR /app
+COPY --from=build /app/bin/Release .
+CMD WcfClient.exe
--- a/samples/wcfapp/Dockerfile.client
+++ b/samples/wcfapp/Dockerfile.client
@@ -1,0 +1,8 @@
+FROM microsoft/dotnet-framework:4.7.2-sdk AS build
+WORKDIR /app
+COPY WcfClient/. ./WcfClient/
+WORKDIR /app/WcfClient
+RUN msbuild
+
+FROM microsoft/dotnet-framework:4.7.2-runtime
+COPY --from=build /app/WcfClient/bin/Debug/ ./
--- a/samples/wcfapp/Dockerfile.client
+++ b/samples/wcfapp/Dockerfile.client
@@ -6,4 +6,4 @@ RUN msbuild /p:Configuration=Release
 FROM microsoft/dotnet-framework:4.7.2-runtime AS runtime
 WORKDIR /app
 COPY --from=build /app/bin/Release .
-CMD WcfClient.exe
+ENTRYPOINT WcfClient.exe
--- a/samples/wcfapp/README.md
+++ b/samples/wcfapp/README.md
@@ -1,0 +1,59 @@
+# WCF Docker Container Samples
+These samples demonstrate how to dockerize WCF services, either IIS-hosted or self-hosted. A simple "hello world" service contract is used in all samples for both HTTP and NET.TCP transport bindings.
+
+## Prepare Your Environment
+After cloning or downloading this repository, build the solution in the `wcfapp` folder either in Visual Studio or using command line `msbuild wcfapp.sln` in a VS console window. You will need environment and software below installed to complete these samples.
+
+* [Windows 10](https://www.microsoft.com/en-us/windows/get-windows-10) (or [Windows Server 2016](https://www.microsoft.com/en-us/cloud-platform/windows-server)) with **'Containers'** feature enabled.
+* [Docker for Windows](https://docs.docker.com/docker-for-windows/) with **'Switch to Windows Containers...'** by right clicking on the Docker whale icon in system tray
+* [Visual Studio 2017](https://www.visualstudio.com/vs/)
+
+## Run WCF Service in Container
+WCF service is supported on .NET Framework, which can run in Windows Server Core based containers. For simplicity, we disabled security in these samples.
+
+### Build a Container with IIS-hosted WCF Service
+Project `WcfServiceWebApp` is created from 'WCF Service Application' template in Visual Studio. A [Dockerfile](WcfServiceWebApp/Dockerfile) is added to the project. We use `microsoft/wcf` image as the base image, which has both HTTP and NET.TCP protocols enabled in IIS and exposes ports 80 (for HTTP) and 808 (for NET.TCP) for the container. We use WCF image with tag `4.6.2` for .NET Framework 4.6.2 in this example, but you can change it to use other tags (eg. `microsoft/wcf:4.7` with tag `4.7`) for WCF images with different versions of .NET Framework. The complete list of [WCF image tags](https://hub.docker.com/r/microsoft/wcf/tags/) can be found from Docker Hub. 
+
+Run commands below to build the container image with name `wcfservice:iis-hosted` and start an instance of it named `myservice1`. Docker parameter `-d` will run the container in background (detached mode).
+```
+C:\wcfapp\WcfServiceWebApp>docker build -t wcfservice:iis-hosted .
+C:\wcfapp\WcfServiceWebApp>docker run -d --name myservice1 wcfservice:iis-hosted
+```
+Find the IP address of the container instance. This will be used later for a WCF client to connect to the service.
+```
+C:\wcfapp\WcfServiceWebApp>docker inspect --format="{{.NetworkSettings.Networks.nat.IPAddress}}" myservice1
+172.23.70.146
+```
+### Build a Container with Self-hosted WCF Service
+Project `WcfServiceConsoleApp` is created from Windows Classic Desktop 'Console App' template in Visual Studio. We added a [Dockerfile](WcfServiceConsoleApp/Dockerfile) to the project. We use `microsoft/dotnet-framework` image as the base image and expose ports 80 (for HTTP) and 808 (for NET.TCP) for the container.
+
+Make sure you build project `WcfServiceConsoleApp` first. Then run commands below to build the container image with name `wcfservice:self-hosted` and start an instance of it named `myservice2`.
+```
+C:\wcfapp\WcfServiceConsoleApp>docker build -t wcfservice:self-hosted .
+C:\wcfapp\WcfServiceConsoleApp>docker run -it --rm --name myservice2 wcfservice:self-hosted
+The service is ready at http://localhost/Service1.svc
+The service is ready at net.tcp://localhost/Service1.svc
+The service is running...
+```
+Open another console window to find the IP address of the self-hosted WCF service (alternatively, you can run the self-hosted WCF service container in detached mode by `docker run -d --name myservice2 wcfservice:self-hosted`).
+```
+C:\wcfapp\WcfClientNetCore>docker inspect --format="{{.NetworkSettings.Networks.nat.IPAddress}}" myservice2
+172.23.69.75
+```
+### Run WCF Client against the Service in Container
+Now that we have WCF services running in containers. Let's run the WCF client against the them. The IP address of the WCF service is passed through the environment variable `host`. *In this example, the WCF client runs on the same machine that hosts the container instance.* 
+```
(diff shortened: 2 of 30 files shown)


samples/wcfapp/WcfClient/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.2" />
```
MichaelSimons: I think it would be best to update the samples to be using the latest FX version - 4.7.2

samples/wcfapp/WcfServiceWebApp/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM microsoft/wcf:4.6.2
+RUN windows\system32\inetsrv\appcmd.exe set app 'Default Web Site/' /enabledProtocols:"http,net.tcp"
+EXPOSE 808
```
MichaelSimons: Is the EXPOSE necessary since it is in the base image?

samples/wcfapp/WcfServiceWebApp/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM microsoft/wcf:4.6.2
+RUN windows\system32\inetsrv\appcmd.exe set app 'Default Web Site/' /enabledProtocols:"http,net.tcp"
```
MichaelSimons: Is this necessary since it is in the base image?

samples/wcfapp/WcfServiceConsoleApp/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM microsoft/dotnet-framework:4.6.2
```
MichaelSimons: What would it take to have a multi-stage Dockerfile that builds the project with one stage and produces the "runtime" image in another?  Can you make use of microsoft/dotnet-framework:4.7.2-sdk to build?  We should be supporting containerized builds.

samples/wcfapp/README.md
```diff
@@ -0,0 +1,59 @@
+# WCF Docker Container Samples
```
MichaelSimons: I would like to see us build and push at least one sample to microsoft/dotnet-framework-samples so that someone can just pull and try something.  e.g. https://github.com/Microsoft/dotnet-framework-docker/blob/master/samples/dotnetapp/README.md#try-a-pre-built-net-framework-docker-image

To do this you can add the appropriate entries in https://github.com/Microsoft/dotnet-framework-docker/blob/master/manifest.samples.json.  I can help you with this if you have questions.

samples/wcfapp/README.md
```diff
@@ -0,0 +1,59 @@
...
+These samples demonstrate how to dockerize WCF services, either IIS-hosted or self-hosted. A simple "hello world" service contract is used in all samples for both HTTP and NET.TCP transport bindings.
+
+## Prepare Your Environment
```
MichaelSimons: I think the best UX would be for the format of this readme to mirror the other samples where it make sense.  As an example, there should be a section that mirrors https://github.com/Microsoft/dotnet-framework-docker/blob/master/samples/dotnetapp/README.md#getting-the-sample.

samples/wcfapp/README.md
```diff
@@ -0,0 +1,59 @@
...
+Hello WCF via Http from Container!
+Hello WCF via Net.Tcp from Container!
+```
+The result above is from running the WCF client against the IIS-hosted WCF service. Changing the environment variable `host` to the IP address of the self-hosted WCF service, you will get a similar result.
+
+*To run WCF client on a different machine other than the one that hosts container instance*, we will need to map ports of a container to ports of its host by using Docker parameter `-p` when we start a constainer instance. For example, the command below started an instance of the self-hosted WCF service container named `myservice3` with its ports 80 and 808 mapped to ports 80 and 808 of its host machine respectively.
+```
+C:\wcfapp\WcfServiceConsoleApp>docker run -d -p 80:80 -p 808:808 --name myservice3 wcfservice:self-hosted
+```
+Then for the WCF client to connect to the service, we need to set the `host` to be the IP address (or DNS name) of the container host machine (instead of the IP address of the container instance). The rest will be the same to start the WCF client.
```
MichaelSimons: At the end, it would be beneficial to link to additional resources - see https://github.com/Microsoft/dotnet-framework-docker/blob/master/samples/dotnetapp/README.md#net-resources.

samples/wcfapp/WcfServiceConsoleApp/Dockerfile
yujayee: @MichaelSimons In this sample, after WCF services are running in the container, users should be able to run WCF client against them. If the projects are build inside container, users still need to build the project locally to have WcfClient.exe. Am I understanding it correctly?

samples/wcfapp/Dockerfile.client
```diff
@@ -0,0 +1,8 @@
+FROM microsoft/dotnet-framework:4.7.2-sdk AS build
+WORKDIR /app
```
MichaelSimons: The nested /app/WcfClient/ path seems unnecessary in the context of Docker.  What do you think about eliminating the nesting?  It could be simplified to just `/app` or `/WcfClient`.

samples/wcfapp/Dockerfile.client
```diff
@@ -0,0 +1,8 @@
...
+COPY WcfClient/. ./WcfClient/
+WORKDIR /app/WcfClient
+RUN msbuild
```
MichaelSimons: I think it would be better to illustrate a production scenario.  In that vein using `/p:Configuration=Release` feels appropriate.

MichaelSimons: You may also choose to specify the output dir as done in https://github.com/Microsoft/dotnet-framework-docker/blob/master/samples/dotnetapp/Dockerfile#L29

samples/wcfapp/Dockerfile.client
```diff
@@ -0,0 +1,8 @@
...
+
+FROM microsoft/dotnet-framework:4.7.2-runtime
+COPY --from=build /app/WcfClient/bin/Debug/ ./
```
MichaelSimons: Any particular reason the app is copied to the root versus an app directory?

samples/wcfapp/README.md
```diff
@@ -0,0 +1,118 @@
...
+Open another console window to find the IP address of the self-hosted WCF service (alternatively, you can run the self-hosted WCF service container in detached mode by `docker run -d --name myservice2 wcfservice:self-hosted`).
+```
+C:\wcfapp\WcfClientNetCore>docker inspect --format="{{.NetworkSettings.Networks.nat.IPAddress}}" myservice2
+172.23.69.75
+```
+### Build a Container to run WCF Client against the Service
+Now that we have WCF services running in containers. Let's run the WCF client against them. Run commands below to build the container image with name `wcfclient`.  
+```
+docker build --pull -t wcfclient -f Dockerfile.client .
+docker run -it --name myclient wcfclient
```
MichaelSimons: Have you considered simplifying these steps by setting the env var with the run command?

`docker run --rm -e HOST=172.23.70.146 wcfclient`

MichaelSimons: Our samples should be encouraging the use of `--rm` to promote cleaning up the Docker resources when done.  I think this usage is a good candidate for using `--rm`.

samples/wcfapp/README.md
```diff
@@ -0,0 +1,118 @@
...
+```
+docker build --pull -t wcfservice:self-hosted -f Dockerfile.console .
+docker run -it --rm --name myservice2 wcfservice:self-hosted
+```
+Open another console window to find the IP address of the self-hosted WCF service (alternatively, you can run the self-hosted WCF service container in detached mode by `docker run -d --name myservice2 wcfservice:self-hosted`).
+```
+C:\wcfapp\WcfClientNetCore>docker inspect --format="{{.NetworkSettings.Networks.nat.IPAddress}}" myservice2
+172.23.69.75
+```
+### Build a Container to run WCF Client against the Service
```
MichaelSimons: What do you think about adding a Docker Compose file that will spin up both the client and service?  It could define a service link and then you can use the host name instead of having to know the IP.  You may need one each for the IIS and self hosted scenarios.

samples/wcfapp/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Build a Container with IIS-hosted WCF Service
+Project `WcfServiceWebApp` is created from 'WCF Service Application' template in Visual Studio. A [Dockerfile](/Dockerfile.web) is added to the project. We use `microsoft/wcf` image as the base image, which has both HTTP and NET.TCP protocols enabled in IIS and exposes ports 80 (for HTTP) and 808 (for NET.TCP) for the container. We use WCF image with tag `4.7.2` for .NET Framework 4.7.2 in this example, but you can change it to use other tags (eg. `microsoft/wcf:4.7` with tag `4.7`) for WCF images with different versions of .NET Framework. The complete list of [WCF image tags](https://hub.docker.com/r/microsoft/wcf/tags/) can be found from Docker Hub. 
+
+Run commands below to build the container image with name `wcfservice:iis-hosted` and start an instance of it named `myservice1`. Docker parameter `-d` will run the container in background (detached mode).
+```
+cd samples
+cd wcfapp
+docker build --pull -t wcfservice:iis-hosted -f Dockerfile.web .
+docker run -d --name myservice1 wcfservice:iis-hosted
```
MichaelSimons: Consider using a more descriptive container name e.g. `iishostedwcfservice`.  Same comment applies to the self-hosted section.

samples/wcfapp/README.md
yujayee: @MichaelSimons I'm not very familiar with Docker-compose, I'll devote some time to this later today.

yujayee:
@MichaelSimons I updated the manifest.samples.json file for WCF prebuilt images, is there anyway I can test these two images?

MichaelSimons:
@yujayee, if you can push your changes to an upstream branch, I can queue a test build for you.

yujayee:
@MichaelSimons Thanks, I just pushed the changes to a new branch wcftestbuild.

MichaelSimons:
@yujayee - I kicked off a [test build](https://devdiv.visualstudio.com/DefaultCollection/DevDiv/_build?_a=summary&buildId=1732727), and pushed the images to https://hub.docker.com/r/msimons/dotnet-framework-samples/tags/

MichaelSimons:
FYI - the 1803 samples didn't build because the branch is out of date.  You'll want to merge in the latest from master at some point.

samples/wcfapp/Dockerfile.client
```diff
@@ -0,0 +1,9 @@
+FROM microsoft/dotnet-framework:4.7.2-sdk AS build
+WORKDIR /app
+COPY WcfClient/. .
+RUN msbuild /p:Configuration=Release
+
+FROM microsoft/dotnet-framework:4.7.2-runtime AS runtime
+WORKDIR /app
+COPY --from=build /app/bin/Release .
+CMD WcfClient.exe
```
MichaelSimons: I wonder if this shouldn't be set as the ENTRYPOINT instead.  It is the only purpose of the image.  Same comment applies to `Dockerfile.console`

samples/wcfapp/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+*To run WCF client on a different machine other than the one that hosts container instance*, we will need to map ports of a container to ports of its host by using Docker parameter `-p` when we start a constainer instance. For example, the command below started an instance of the self-hosted WCF service container named `myservice3` with its ports 80 and 808 mapped to ports 80 and 808 of its host machine respectively.
+```
+C:\wcfapp\WcfServiceConsoleApp>docker run -d -p 80:80 -p 808:808 --name myservice3 selfhostedwcfservice
+```
+Then for the WCF client to connect to the service, we need to set the `host` to be the IP address (or DNS name) of the container host machine (instead of the IP address of the container instance). The rest will be the same to start the WCF client.
+
+## Build and run the sample with Docker Compose
+[Docker Compose](https://docs.docker.com/compose/overview/) is a tool for defining and running multi-container Docker applications. In this sample, we also added YML files to configure WCF server and client applications. You can directly start and run service and client applications without finding IP address of service container.
```
MichaelSimons: What do you think about changing `without finding IP address of service container` to `without having to any work to hookup the client to the WCF service"

MichaelSimons:
@yujayee - As soon as I merged, I realized one more thing needs to be updated.  Can you update the [readme](https://github.com/Microsoft/dotnet-framework-docker/blob/master/samples/README.DockerHub.md) posted to [microsoft/dotnet-framework-samples](https://hub.docker.com/r/microsoft/dotnet-framework-samples/)?   Just create a PR to update the readme and once merged, the build infra will automatically update the readme on DockerHub.  Thanks

PR #197, opened 2018-11-13
11 b updates

MichaelSimons:
CI is expected to fail until the updated windows base images are available.


MichaelSimons:
cc @rotata - FYI

Issue #228, opened 2019-03-08
Update SDK images to VS 2019


MichaelSimons:
Currently the SDK images contain the VS 2017 Test Agent and Build Tools.  Once VS 2019 is released, the SDK images should be upgraded.

PR #252, opened 2019-04-01
Update to reference VS 2019 build tools.

mthalman:
Updates the SDK Dockerfiles to be prepared for referencing the VS 2019 build tools.  This basically consists of updates to paths and some ngen workarounds.  Placeholders have been added for the URLs where the final assets will be downloaded from.

Also includes some changes to delete the VS Packages folder for SDK images and the TEMP folder for runtime images to save on space of the image.

windowservercore-1709 images have not been updated since that is going to be EOL and want to avoid any potential breakage there.

### DIFF
--- a/3.5/runtime/windowsservercore-1803/Dockerfile
+++ b/3.5/runtime/windowsservercore-1803/Dockerfile
@@ -13,8 +13,7 @@ RUN powershell -Command `
         Expand-Archive microsoft-windows-netfx3.zip; `
         Remove-Item -Force microsoft-windows-netfx3.zip; `
         Add-WindowsPackage -Online -PackagePath .\microsoft-windows-netfx3\microsoft-windows-netfx3-ondemand-package~31bf3856ad364e35~amd64~~.cab; `
-        Remove-Item -Force -Recurse microsoft-windows-netfx3; `
-        Remove-Item -Force -Recurse ${Env:TEMP}\*
+        Remove-Item -Force -Recurse microsoft-windows-netfx3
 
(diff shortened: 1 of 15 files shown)


3.5/runtime/windowsservercore-1803/Dockerfile
```diff
@@ -13,7 +13,8 @@ RUN powershell -Command `
         Expand-Archive microsoft-windows-netfx3.zip; `
         Remove-Item -Force microsoft-windows-netfx3.zip; `
         Add-WindowsPackage -Online -PackagePath .\microsoft-windows-netfx3\microsoft-windows-netfx3-ondemand-package~31bf3856ad364e35~amd64~~.cab; `
-        Remove-Item -Force -Recurse microsoft-windows-netfx3
```
MichaelSimons: I think we should defer rolling these cleanup changes within the runtiem Dockerfiles until the next time the runtime images are patched.  We don't want to rebuild the runtime image with this VS2019 SDK update.

manifest.json
```diff
@@ -3,7 +3,7 @@
   "registry": "mcr.microsoft.com",
   "variables": {
     "RuntimeReleaseDateStamp": "20190312",
-    "SdkReleaseDateStamp": "TODO_SDK_DATESTAMP"
+    "SdkReleaseDateStamp": "20190402"
```
MichaelSimons: Please run Get-TagsDocumentation.ps1 to update the tags documentation.

mthalman:
Fixes #228

PR #257, opened 2019-04-05
Delete temp folder after installing .NET Fx

### LINKED ISSUE #255: Delete temp files from runtime images
The act of installing the .NET Fx Runtime will produce log files in the %TEMP% folder.  These should be cleaned up in the image.

mthalman:
.NET Fx leaves files in the TEMP folder after running the installer.  These should be cleaned up.

Fixes #255

### DIFF
--- a/4.7/runtime/windowsservercore-ltsc2016/Dockerfile
+++ b/4.7/runtime/windowsservercore-ltsc2016/Dockerfile
@@ -5,7 +5,8 @@ FROM mcr.microsoft.com/windows/servercore:ltsc2016
 # Install .NET 4.7
 RUN powershell Invoke-WebRequest -Uri "https://download.microsoft.com/download/D/D/3/DD35CC25-6E9C-484B-A746-C5BE0C923290/NDP47-KB3186497-x86-x64-AllOS-ENU.exe" -OutFile dotnet-framework-installer.exe & `
     .\dotnet-framework-installer.exe /q & `
-    del .\dotnet-framework-installer.exe
+    del .\dotnet-framework-installer.exe & `
+    powershell Remove-Item -Force -Recurse ${Env:TEMP}\*
 
 # Apply latest patch
 RUN powershell -Command `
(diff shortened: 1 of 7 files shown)


4.7/runtime/windowsservercore-ltsc2016/Dockerfile
```diff
@@ -5,7 +5,8 @@ FROM mcr.microsoft.com/windows/servercore:ltsc2016
 # Install .NET 4.7
 RUN powershell Invoke-WebRequest -Uri "https://download.microsoft.com/download/D/D/3/DD35CC25-6E9C-484B-A746-C5BE0C923290/NDP47-KB3186497-x86-x64-AllOS-ENU.exe" -OutFile dotnet-framework-installer.exe & `
     .\dotnet-framework-installer.exe /q & `
-    del .\dotnet-framework-installer.exe
+    del .\dotnet-framework-installer.exe & `
+    powershell Remove-Item -Force -Recurse ${Env:TEMP}\*
```
MichaelSimons: For these ltsc2016 Dockerfiles, it feels like `del /S /F /AH %TEMP%` would be better since the RUN instructions are being executed in a CMD shell.  Is there a reason to use powershell?

mthalman: The `del` command doesn't delete nested folders.

MichaelSimons: `rmdir` then or a combination of both?  I ask because of the desire to not be using a mixture of shells unless required.

mthalman: I have not come across a reasonable solution to this without using PowerShell.  Examples of hacks or batch files that accomplish this with basic command shell can be found here: https://superuser.com/questions/173859/how-can-i-delete-all-files-subfolders-in-a-given-folder-via-the-command-prompt.  The problem is that we want to delete the contents of %TEMP% but not the actual %TEMP% folder.  This is what PowerShell provides in a simple command.

PR #258, opened 2019-04-05
Reduce usage of SHELL command

### LINKED ISSUE #251: Avoid switching between shells in Dockerfiles
The Dockerfiles make use of the `SHELL` command to switch between either cmd or PowerShell.  Each `SHELL` command adds a layer to the image so we should avoid using it when possible.  However, all SDK images should set PowerShell as the default shell, so there needs to be at least one command to do that.

mthalman:
Using the SHELL command adds a layer to the image, so we should reduce the usage to only those that are necessary.  This is done by using the default shell, `cmd`, throughout the Dockerfile until we only have Powershell commands to execute (we want the image to use Powershell as the default shell) at which point we use `SHELL` to switch to Powershell.

Fixed #251.

### DIFF
--- a/3.5/sdk/windowsservercore-1803/Dockerfile
+++ b/3.5/sdk/windowsservercore-1803/Dockerfile
@@ -3,44 +3,49 @@
 ARG REPO=mcr.microsoft.com/dotnet/framework/runtime
 FROM $REPO:3.5-windowsservercore-1803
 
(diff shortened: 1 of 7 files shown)


mthalman:
@dotnet-bot test WindowsServerCore-1803 - 3.5 Dockerfiles please

mthalman:
Closing this without merging in order to do a more significant refactoring of how commands are being run in the Dockerfiles.

Issue #265, opened 2019-04-10
vswhere doesn't work after clean up


Smenus:
After a recent addition to the clean up after installing VS2019 build tools, it is no longer possible to use vswhere to find the VS installation.

Running the following inside an interactive powershell using sdk:4.7.2 results in an empty array
```ps
Invoke-WebRequest -UseBasicParsing https://github.com/Microsoft/vswhere/releases/download/2.6.7/vswhere.exe -OutFile vswhere.exe
.\vswhere.exe -version "[16.0,17.0)" -products Microsoft.VisualStudio.Product.BuildTools -latest -format json
```

I've made a copy of the 4.7.2 Dockerfile, and when I comment out `Remove-Item -Force -Recurse \"${Env:ProgramData}\Microsoft\VisualStudio\Packages\";`, it shows the Build Tools installation.

MichaelSimons:
@mthalman - please investigate.

heaths:
You can [delete the package cache](https://github.com/microsoft/dotnet-framework-docker/blob/master/4.7.2/sdk/windowsservercore-ltsc2019/Dockerfile#L32) but should exclude "_Instances":

```powershell
Get-ChildItem \"${Env:ProgramData}\Microsoft\VisualStudio\Packages\" -Include \"*\" -Exclude \"_Instances\" | Remove-Item -Force -Recurse
```

vswhere.exe and potentially other tools in Build Tools need this to function properly.

heaths:
Closed without reason or resolution. Why? As one of the developers on the setup engine and who wrote the query API, I can assure you this is a problem. Do not delete the instance directory. It's not that big.

heaths:
FYI: you can link commits and PRs too bug and even automatically close them by adding in the description, "Fixes #1234".

Thanks.

mattemoore:
This issue still exists as https://github.com/microsoft/dotnet-framework-docker/pull/267 removes the removal of the `packages` folder but the removal of the `installer` folder is still done, thus `vswhere` is still being removed.

mthalman:
@mattemoore - This issue is fixed because it was about `vswhere.exe` not working due to the removal of the `Packages` folder, but required that `vswhere.exe` be manually copied into the container.  Having `vswhere.exe` already exist in the published image is a separate issue and you're encouraged to log an issue for it if that's something you want to have.

PR #267, opened 2019-04-11
Ensure that VS packages folder is not deleted as part of cleanup

mthalman:
It was originally thought that the VS packages folder could be deleted from the image to provide some savings on image size.  But that folder is actually needed in order to execute any VS installer commands.  For example, modifying the existing installation with an additional component.  Without the VS packages folder, the installer will not recognize VS as already being installed.  So we should revert the change to remove this folder.


heaths:
This still leaves a lot of cruft behind. My suggestion on #265 would reduce image size while maintaining the state information required.

PR #277, opened 2019-04-23
Refactor pipeline for PR builds that can build/test in same job

mthalman:
This is basically just a cherry-pick of [4cfc39e](https://github.com/dotnet/dotnet-docker/commit/4cfc39eff26f67786245972625e22d9e810806d2).  It's been adapted for this repo by updating the manifest to include testing dependencies for certain sections of the graph.  The build job timeout has been increased due to the number of images that need to be built in one job.


.vsts-pipelines/variables/docker-images.yml
```diff
@@ -1,3 +1,3 @@
 variables:
-  imageNames.imageBuilder.linux: mcr.microsoft.com/dotnet-buildtools/image-builder:debian-20190326143958
-  imageNames.imageBuilder.windows: mcr.microsoft.com/dotnet-buildtools/image-builder:nanoserver-20190326144001
+  imageNames.imageBuilder.linux: mcr.microsoft.com/dotnet-buildtools/image-builder:debian-20190410133145
```
MichaelSimons: Would you mind updating this to the latest `20190420013239` to get the publish manifest optimization?

.vsts-pipelines/steps/test-images-linux-client.yml
```diff
@@ -0,0 +1,77 @@
+parameters:
```
MichaelSimons: I think this should be deleted as it isn't applicable here.

manifest.json
```diff
@@ -90,7 +110,18 @@
               "tags": {
                 "4.6.2-$(RuntimeReleaseDateStamp)-windowsservercore-ltsc2016": {},
                 "4.6.2-windowsservercore-ltsc2016": {}
-              }
+              },
+              "customBuildLegGrouping": [
+                {
+                  "name": "pr-build",
+                  "dependencies": [
+                    "$(Repo:runtime):4.7.1-windowsservercore-ltsc2016",
```
MichaelSimons: Where is the dependency on the 4.7.1 runtime coming from?  Isn't the dependency just on the sdk?  This comment applies to other groupings as well.

manifest.json
```diff
@@ -58,7 +58,16 @@
               "tags": {
                 "4.7.1-$(RuntimeReleaseDateStamp)-windowsservercore-ltsc2016": {},
                 "4.7.1-windowsservercore-ltsc2016": {}
-              }
+              },
```
MichaelSimons: Friendly reminder to pull in latest and the changes that added 4.8.  You will need to declare dependencies on 4.8 as well right?

manifest.json
mthalman: The dependency graph needs to be flattened because of how Image Builder works.  Ideally, we should just add a dependency on the SDK and get all of it's dependencies along with it, but that's not possible with the current logic.  I'll log a bug on Image Builder.

mthalman: https://github.com/dotnet/docker-tools/issues/175

.vsts-pipelines/steps/test-images-linux-client.yml
MichaelSimons: What was the resolution to this?  I realize now my comment wasn't very clear.  I was referring to the entire yml file.  It isn't used at all therefore I think it should be removed to eliminate confusion.  I realize at some point in the future if we share the build infra it may get copied here but until that happens, I think it should not be added.

mthalman: Sorry, I wasn't thinking. I thought you were referring to the parameters not being needed, not the entire file.  But yes, this entire file can go.

PR #287, opened 2019-05-20
Update timestamp to get re-released Windows base image

mthalman:
Windows re-released an update for last week's patch Tuesday.  Updating the timestamps so we can release updated images.


mthalman:
Aborting this change.  Will update images without change the timestamp.

PR #296, opened 2019-06-11
Updates for Patch Tuesday

mthalman:


3.5/runtime/windowsservercore-1903/Dockerfile
```diff
@@ -10,6 +10,14 @@ RUN curl -fSLo microsoft-windows-netfx3.zip https://dotnetbinaries.blob.core.win
     && del microsoft-windows-netfx3-ondemand-package~31bf3856ad364e35~amd64~~.cab `
     && powershell Remove-Item -Force -Recurse ${Env:TEMP}\*
 
+# Apply latest patch
```
MichaelSimons: I'm not following why this is necessary.  Isn't this patch the OS LCU?  .NET for 1903 is patched separately from the OS.

mthalman: Ah, yep.

manifest.json
```diff
@@ -2,8 +2,8 @@
   "readmePath": "README.md",
   "registry": "mcr.microsoft.com",
   "variables": {
-    "RuntimeReleaseDateStamp": "20190514",
-    "SdkReleaseDateStamp": "20190521"
+    "RuntimeReleaseDateStamp": "20190611",
```
MichaelSimons: The repo readme's should be regenerated as a result of this change.

mthalman: Yep, just committed it.

PR #309, opened 2019-07-03
Clean up sdk_tools48.cab file

### LINKED ISSUE #297: SDK images contain sdk_tools48.cab file
As part of the installation of .NET 4.8 SDK, there's a sdk_tools48.cab file leftover in the image, contained in the root directory.  This needs to be deleted as part of the Dockerfile definition.

mthalman:
When extracting the sdk_tools48.zip file, both a .cab and .msi file are placed in the directory but only the .msi file is being cleaned up.  These changes ensure that everything is properly cleaned up.

Fixes #297

### DIFF
--- a/3.5/sdk/windowsservercore-1803/Dockerfile
+++ b/3.5/sdk/windowsservercore-1803/Dockerfile
@@ -43,8 +43,8 @@ RUN `
 
 # Install .NET 4.8 SDK
 RUN mkdir sdk_tools48 `
-    && curl -fSLo sdk_tools48/sdk_tools48.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/sdk_tools48.zip `
-    && tar -zxf sdk_tools48/sdk_tools48.zip -C sdk_tools48 `
+    && curl -fSLo sdk_tools48\sdk_tools48.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/sdk_tools48.zip `
+    && tar -zxf sdk_tools48\sdk_tools48.zip -C sdk_tools48 `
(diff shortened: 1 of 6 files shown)


3.5/sdk/windowsservercore-1803/Dockerfile
```diff
@@ -45,8 +45,7 @@ RUN `
 RUN curl -fSLo sdk_tools48.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/sdk_tools48.zip `
     && tar -zxf sdk_tools48.zip `
```
MichaelSimons: Thoughts on utilizing a temp directory to extract the contents to and then simply removing the directory?  It feels like that is a safer more robust pattern.

3.5/sdk/windowsservercore-1803/Dockerfile
```diff
@@ -42,11 +42,11 @@ RUN `
     && rmdir /S /Q "%ProgramData%\Package Cache"
 
 # Install .NET 4.8 SDK
-RUN curl -fSLo sdk_tools48.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/sdk_tools48.zip `
-    && tar -zxf sdk_tools48.zip `
-    && start /w msiexec /i sdk_tools48.msi /quiet VSEXTUI=1 `
-    && del sdk_tools48.msi `
-    && del sdk_tools48.zip `
+RUN mkdir sdk_tools48 `
+    && curl -fSLo sdk_tools48/sdk_tools48.zip https://dotnetbinaries.blob.core.windows.net/dockerassets/sdk_tools48.zip `
```
MichaelSimons: From a consistency standpoint, would it make sense to use `\` in the paths consistently?  e.g. line 46 and 47?

Issue #302, opened 2019-06-18
Migrate WCF Dockerfiles and image production into this repo


MichaelSimons:
**Reasons**
1. Provides source code alignment with the [.NET Fx docker product family repo](https://hub.docker.com/_/microsoft-dotnet-framework/)
1. Customers are logging issues for the various .NET docker images within the wrong repos from time to time.  This consolidation eliminates the confusion on which repo to log issues with.
1. All of the .NET Fx Samples are already located within this repo.  Having all of the .NET Fx Dockerfiles used by the samples in the same location has value.
1. Having all of the .NET Fx Dockerfiles within a single repo eliminate overhead in sharing the same engineering infrastructure.

**Tasks**
- [x] Migrate Dockerfiles and readmes from https://github.com/Microsoft/wcf-docker to this repo.
- [x] Define manifest.wcf.json which describes the WCF Dockerfiles and the tags to use on the resulting images.
- [x] Define a WCF build definition.  I believe it will mimic the [Samples build](https://github.com/microsoft/dotnet-framework-docker/blob/master/.vsts-pipelines/builds/dotnet-framework-samples.yml) closely
- [x] Migrate the MCR WCF repo definition from iis to dotnet
- [x] Remove the WCF readme and tags metadata from the iis mcrdocs.  I believe the appropriate dotnet docs will get automatically created via the .NET build infrastructure.
- [x] Create the appropriate automated tests hooked up to the builds
- [x] Copy all open https://github.com/microsoft/wcf-docker issues to https://github.com/microsoft/dotnet-framework-docker
- [x] Archive the https://github.com/microsoft/wcf-docker repo
~~- [ ] Hookup WCF builds to docker-tools base image change detection system~~

MichaelSimons:
This work has been completed.  Closing.

PR #317, opened 2019-07-10
Initial checkin for ASP.NET docker repo migration

HongGit:
Please don't merge yet!!!!!

This is the initial checkin for issue #303, This change includes:

- Dockerfile migration
- manifest.aspnet.json and aspnet-tages.yml

PR #321, opened 2019-07-15
Add the yml for the WCF pipeline.

StephenBonikowsky:
@MichaelSimons For step...

> Define a WCF build definition. I believe it will mimic the Samples build closely

of #302 

I just need to add this yml and then create a new build definition called "dotnet-framework-docker-wcf" that maps to this yml and add it to...
 'dnceng/internal/Pipelines/Builds' under Microsoft/dotnet-framework-docker.

Am I missing anything else?

### DIFF
--- a/eng/pipelines/dotnet-framework-wcf.yml
+++ b/eng/pipelines/dotnet-framework-wcf.yml
@@ -1,0 +1,14 @@
+trigger: none
+pr:
+- master
+
+variables:
+- template: variables/common.yml
+- name: manifest
+  value: manifest.wcf.json
+stages:
+- template: ../common/templates/stages/build-test-publish-repo.yml
+  parameters:
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      buildMatrixType: platformVersionedOs
+      customBuildLegGrouping: pr-build


eng/pipelines/dotnet-framework-wcf.yml
```diff
@@ -0,0 +1,12 @@
+trigger: none
+pr: none
```
MichaelSimons: I think this should have a pr trigger on master.  e.g. https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/pipelines/dotnet-framework.yml#L2

MichaelSimons:
> I just need to add this yml and then create a new build definition called "dotnet-framework-docker-wcf" that maps to this yml and add it to...
> 'dnceng/internal/Pipelines/Builds' under Microsoft/dotnet-framework-docker.

That is correct.  You will also want to add a build to PR validation in 'dnceng/public/Pipelines/Builds'

How are these changes getting coordinated with the active PRs @HongGit has open (https://github.com/microsoft/dotnet-framework-docker/pull/318 and https://github.com/microsoft/dotnet-framework-docker/pull/317)?

eng/pipelines/dotnet-framework-wcf.yml
MichaelSimons: Can/should path filters be specified to only trigger PR validation when WCF artifacts are changes?

StephenBonikowsky:
I could push these changes to her branch so they become part of her PR or I can merge after her's goes in.

eng/pipelines/dotnet-framework-wcf.yml
StephenBonikowsky: done

MichaelSimons:
I think combining with @HongGit work is a good idea.  I would like to see the e2e validated.  Creating a feature branch may be a good way to make this easier to manage.  That will allow more incremental reviews/checkins, support multiple contributors, and allow e2e validation prior to merging to master.

eng/pipelines/dotnet-framework-wcf.yml
mthalman: I'm not sure we should use path triggers for PRs.  If we make changes to the pipeline infrastructure, I would want this pipeline to be validated for those changes.

MichaelSimons: Multiple trigger paths are a possibility... The primary reason I mention this is that if a WCF only PR is made, I wouldn't want it to wait for the long running SDK tests to complete in order for it to turn green.  This type of scenario could waste a lot of compute.

That being said, I am fine with "tuning" these scenarios after the initial onboarding is complete.

StephenBonikowsky:
Ok, I'm pushing these changes to @HongGit 's branch (#322) and closing this PR.

PR #325, opened 2019-07-17
Minor fixes to the wcf manifest file.

StephenBonikowsky:


StephenBonikowsky:
@mthalman @MichaelSimons 
With these changes we are now building the images just fine.
https://dev.azure.com/dnceng/internal/_build/results?buildId=269645

PR #330, opened 2019-07-23
Make TestCategory opt-in.

StephenBonikowsky:
Change the logic in the infra scripts so that the new test filtering argument doesn't impact any existing tests and any future repo could opt-in to it if they wish.


StephenBonikowsky:
@MichaelSimons Is this more like what you were requesting by your comments in https://github.com/microsoft/dotnet-framework-docker/pull/318#discussion_r306406195

eng/common/templates/steps/test-images-windows-client.yml
```diff
@@ -1,9 +1,12 @@
 steps:
 - powershell: |
     if ("${{ eq(variables['System.TeamProject'], 'public') }}" -eq "False") {
-      $optionalTestArgs="-Registry $env:ACR_SERVER -RepoPrefix $env:STAGINGREPOPREFIX -TestCategory $env:TESTCATEGORY"
+      $optionalTestArgs="-Registry $env:ACR_SERVER -RepoPrefix $env:STAGINGREPOPREFIX"
     } else {
-      $optionalTestArgs="-IsLocalRun -TestCategory $env:TESTCATEGORY"
+      $optionalTestArgs="-IsLocalRun"
+    }
+    if ($repoTestArgs) {
```
MichaelSimons: Does this need to be using `$env:REPOTESTARGS`?

eng/common/templates/steps/test-images-windows-client.yml
```diff
@@ -1,9 +1,12 @@
...
+      $optionalTestArgs=$optionalTestArgs + "-TestCategory $env:REPOTESTARGS"
```
MichaelSimons: I would expect this to be: 

```suggestion
      $optionalTestArgs += " $env:REPOTESTARGS"
```

The important difference is that the other repos that are utilizing the **_common_** infrastructure should be able to use the $env:REPOTESTARGS to specify a completely different argument to the test script.

eng/pipelines/dotnet-framework-wcf.yml
```diff
@@ -6,8 +6,8 @@ variables:
 - template: variables/common.yml
 - name: manifest
   value: manifest.wcf.json
-- name: testCategory
-  value: WCF
+- name: repoTestArgs 
+  value: "--filter -Category WCF"
```
MichaelSimons: I wouldn't expect `--filter` to be specified here.  That could be encapsulated within RunTests.

eng/pipelines/dotnet-framework.yml
```diff
@@ -6,8 +6,6 @@ variables:
 - template: variables/common.yml
 - name: manifest
   value: manifest.json
-- name: testCategory
```
MichaelSimons: I would expect a version of this to remain.

```
- name: repoTestArgs 
  value: "-Category Runtime"
```

tests/run-tests.ps1
```diff
@@ -42,5 +42,10 @@ if ($IsLocalRun) {
     $env:LOCAL_RUN = 1
 }
 
-& dotnet test --filter Category=$TestCategory -c Release --logger:trx $PSScriptRoot/Microsoft.DotNet.Framework.Docker.Tests/Microsoft.DotNet.Framework.Docker.Tests.csproj
+if ($TestCategory) {
+    & dotnet test $TestCategory -c Release --logger:trx $PSScriptRoot/Microsoft.DotNet.Framework.Docker.Tests/Microsoft.DotNet.Framework.Docker.Tests.csproj
+} else  {
+  & dotnet test -c Release --logger:trx $PSScriptRoot/Microsoft.DotNet.Framework.Docker.Tests/Microsoft.DotNet.Framework.Docker.Tests.csproj
+}
```
mthalman: Should be able to dynamically construct the parameters to `dotnet test` without duplicating them and then just have one spot that calls `dotnet test`.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -76,15 +76,13 @@ private static string GetFilterRegexPattern(string filter)
         }
 
         [Theory]
-        [Trait("Category", "Runtime")]
```
MichaelSimons: I was expecting this trait to remain.

eng/common/templates/steps/test-images-windows-client.yml
mthalman:
```suggestion
      $optionalTestArgs=$optionalTestArgs + " -TestCategory $env:REPOTESTARGS"
```

MichaelSimons:
Is there a need for this second PR?  Can we keep the changes in #318?  IMO makes it easier to manage/track/resolve the review comments.

StephenBonikowsky:
I can push directly to hongli/wcf.

eng/common/templates/steps/test-images-windows-client.yml
StephenBonikowsky: Possibly, not very familiar with PS, I'll test it in azdo to confirm.

tests/run-tests.ps1
StephenBonikowsky: Based on Michael's additional feedback I think this can be reverted to how it was.

eng/common/templates/steps/test-images-windows-client.yml
StephenBonikowsky: Understood.

StephenBonikowsky:
I just pushed these changes plus edits based on your feedback directly to #318 so we can now close this PR.

PR #346, opened 2019-08-09
Add build-and-test script for usage in local dev workflow

### LINKED ISSUE #340: The dev workflow for building and testing images regressed with the introduction of the shared Docker engineering infra
1. There is no longer a script at the root of the repo to run a build and test.
1. The https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/common/build-and-test.ps1 script in the common engineering folder doesn't work in this repo because there are no architecture folders.  The script also will need to be modified with the addition of wcf and aspnet because they have their own manifest files.

MichaelSimons:
Fixes #340 

`.\build-and-test.ps1 -OSFilter *1903 -VersionFilter 4.8` --> builds all runtime and sdk 4.8 images for windowsservercore-1903
`.\build-and-test.ps1 -manifestType wcf -OSFilter *1903 -VersionFilter 4.8` --> builds all wcf 4.8 images for windowsservercore-1903


MichaelSimons:
FYI @HongGit, @StephenBonikowsky

.dockerignore
```diff
@@ -0,0 +1,8 @@
+**/bin
+**/obj
+**/out
+**/.vscode
+**/.vs
+.dotnet
+.Microsoft.DotNet.ImageBuilder
```
mthalman: Not needed

build-and-test.ps1
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env pwsh
+[cmdletbinding()]
+param(
+    [ValidateSet('aspnet','wcf','samples')]
```
mthalman: For consistency, it would make sense to list "runtime" here but continue to default to "runtime" if it's not set.

build-and-test.ps1
```diff
@@ -0,0 +1,29 @@
...
+    [string]$ManifestType,
+    [string]$VersionFilter = "*",
+    [string]$OSFilter = "*",
+    [string]$OptionalImageBuilderArgs,
+    [switch]$SkipTesting = $false
+)
+
+$manifestTypeExt = $ManifestType
+if (-not [string]::IsNullOrEmpty($manifestType)) {
```
mthalman: This can be simplified to `if ($manifestType) {`

.dockerignore
MichaelSimons: It is extremely helpful when running `Get-TagsDocumentation`

build-and-test.ps1
MichaelSimons: I debated a long time on this.  What do you think about renaming `manifest.json` to `manifest.runtime.json`?  The reason is there isn't a clear default anymore.

MichaelSimons: I think it would help make things more clear and consistent.

mthalman: Good point.  Although now that I think about it, the "runtime" terminology isn't very fitting considering it also contains metadata for the SDK images.  What about "base"?

MichaelSimons: My vote would be something more descriptive maybe - `manifest.runtime.sdk.json`  In this context `base` isn't meaningful to me.

mthalman: How about a very slight tweak: `manifest.runtime-sdk.json`

MichaelSimons: Done

build-and-test.ps1
MichaelSimons: This is no longer applicable after the `manifest.runtime-sdk.json` refactoring.

MichaelSimons: To be more specific, the code was removed.

mthalman:
What do you think about renaming the dotnet-framework.yml pipeline file to dotnet-framework-runtime-sdk.yml?

MichaelSimons:
@mthalman  - yes I'll rename the pipeline.

PR #382, opened 2019-08-19
Integrate eng/common folder

mthalman:
Integrates the latest changes in the eng/common folder.

Since the logic in eng/common folder is now passing the ImageInfoPath argument to the run-tests PowerShell script, I've updated the script and test code to honor that argument and to pull images from the appropriate location (staging or MCR) based on whether the image tag is contained in the image info file.

### DIFF
--- a/tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
@@ -265,7 +266,7 @@ namespace Microsoft.DotNet.Framework.Docker.Tests
         {
             string repo = $"dotnet/framework/{imageType}";
             string tag = $"{version}-{osVariant}";
-            string registry = GetRegistryName(repo, tag);
+            string registry = GetRegistryNameWithRepoPrefix(repo, tag);
             string image = $"{registry}{repo}:{tag}";
 
             // Ensure image exists locally
@@ -281,12 +282,12 @@ namespace Microsoft.DotNet.Framework.Docker.Tests
             return image;
         }
 
-        private static string GetRegistryName(string repo, string tag)
+        private static string GetRegistryNameWithRepoPrefix(string repo, string tag)
         {
-            bool imageExistsInStaging = true;
+            bool isUsingCustomRegistry = true;
 
             // In the case of running this in a local development environment, there would likely be no image info file
-            // provided. In that case, the assumption is that the images exist in the staging location.
+            // provided. In that case, the assumption is that the images exist in the custom configured location.
 
             if (ImageTests.ImageInfoData.Value != null)
             {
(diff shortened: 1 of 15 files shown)


tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -260,6 +281,33 @@ private string GetImage(string imageType, string version, string osVariant)
             return image;
         }
 
+        private static string GetRegistryName(string repo, string tag)
+        {
+            bool imageExistsInStaging = true;
+
+            // In the case of running this in a local development environment, there would likely be no image info file
+            // provided. In that case, the assumption is that the images exist in the staging location.
```
MichaelSimons: > In that case, the assumption is that the images exist in the staging location.

This comment as well as the implementation doesn't seem correct.  In a local dev env, a staging location is not being used.  This would not be desirable behavior IMO.  If you build locally, you should end up with the mcr based tags.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -260,6 +281,33 @@ private string GetImage(string imageType, string version, string osVariant)
             return image;
         }
 
+        private static string GetRegistryName(string repo, string tag)
```
MichaelSimons: Technically, this name isn't correct because in the case of the staging location, it is a registry plus a repo prefix.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -260,6 +281,33 @@ private string GetImage(string imageType, string version, string osVariant)
...
+                        .Cast<JProperty>()
+                        .Any(imageInfo => imageInfo.Value["simpleTags"].Any(imageTag => imageTag.ToString() == tag));
+                }
+                else
+                {
+                    imageExistsInStaging = false;
+                }
+            }
+
+            return imageExistsInStaging ? $"{Registry}/{RepoPrefix}" : "mcr.microsoft.com/";
```
MichaelSimons: Would it make sense to assert/validate that `Registry` and `RepoPrefix` have been set in the staging code path?

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
mthalman: How did it work before?  Before my changes, it _never_ went to MCR.  So was the dev environment before incorrect before?

MichaelSimons: It looks like `$"{Registry}/{RepoPrefix}"` resolves to `mcr.microsoft.com` in the dev flow because it the environment variables aren't being specified in which it retrieves the registry from the manifest.  So the proposal seem functional but not ideal in that it is not using the staging location in reality.

mthalman: Maybe instead of using "staging" terminology, I'll just use "custom".

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
mthalman: Renamed to `GetRegistryNameWithRepoPrefix`.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -295,17 +296,24 @@ private static string GetRegistryName(string repo, string tag)
...
                 }
                 else
                 {
-                    imageExistsInStaging = false;
+                    isUsingCustomRegistry = false;
                 }
             }
 
-            return imageExistsInStaging ? $"{Registry}/{RepoPrefix}" : "mcr.microsoft.com/";
+            if (isUsingCustomRegistry)
```
MichaelSimons: After looking at the implementation more closely, I'm not certain these asserts are valid.  The registry gets defaulted to mcr.microsoft.com if the env var is not specified.  Also it is valid to have an empty Prefix.  I think the original pattern was better.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -295,17 +296,24 @@ private static string GetRegistryName(string repo, string tag)
...
+            {
+                Debug.Assert(!String.IsNullOrEmpty(Registry), "Registry is not set.");
+                Debug.Assert(!String.IsNullOrEmpty(RepoPrefix), "RepoPrefix is not set.");
+                return $"{Registry}/{RepoPrefix}";
+            }
+            
+            return "mcr.microsoft.com/";
```
MichaelSimons: I think the hard coded MCR registry reference should be changed to GetManifestRegistry().  e.g. this should only be specified in one place within the source.

PR #384, opened 2019-08-22
Combine ASP.NET and WCF with runtime/SDK into one build

mthalman:


tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -161,19 +164,21 @@ public void VerifyImagesWithWebApps(ImageDescriptor imageDescriptor)
...
 
-        [Theory]
+        [SkippableTheory("3.5")]
+        [Trait("Category", "all")]
         [Trait("Category", "WCF")]
         [MemberData(nameof(GetVerifyWcfImagesData))]
         public void VerifyWcfImagesWithApps(ImageDescriptor imageDescriptor)
         {
-            VerifyWcfImages(imageDescriptor);
+            //VerifyWcfImages(imageDescriptor);
```
MichaelSimons: Test change?  It looks like this should be reverted.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -72,6 +72,7 @@ public class ImageTests
 
         private static ImageDescriptor[] WcfTestData = new ImageDescriptor[]
         {
+            new ImageDescriptor { RuntimeVersion = "3.5", OsVariant = "" }, // Placeholder to avoid empty MemberData for 3.5 (see https://github.com/xunit/xunit/issues/1113)
```
MichaelSimons: long line

build-and-test.ps1
```diff
@@ -1,20 +1,26 @@
 #!/usr/bin/env pwsh
 [cmdletbinding()]
 param(
-    [ValidateSet('runtime-sdk', 'aspnet', 'wcf', 'samples')]
-    [string]$ManifestType = 'runtime-sdk',
+    [ValidateSet('default', 'samples')]
```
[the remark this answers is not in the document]
mthalman: Fixed.  Required some changes to the common script because we need to use multiple path filters to support both runtime and SDK images.

I also removed `samples` because the build for that doesn't work in dev environment due to their being multiple Dockerfiles with the same build context path that are attempted to be built.  This causes an exception in the BuildCommand.  This doesn't happen on official builds because of path filtering to only build for a specific OS.

build-and-test.ps1
```diff
@@ -1,26 +1,30 @@
 #!/usr/bin/env pwsh
 [cmdletbinding()]
 param(
-    [ValidateSet('default', 'samples')]
-    [string]$ManifestType = 'default',
+    [ValidateSet('*', 'runtime-sdk', 'aspnet', 'wcf')]
+    [string]$ImageTypeFilter = '*',
```
MichaelSimons: It feels more natural and efficient to me if this were using the `repo` filtering capabilities of ImageBuilder.  I also think renaming this parameter in terms of repo would be more natural and meaningful to users.  You could also easily support multiples by making this an array.

build-and-test.ps1
```diff
@@ -1,26 +1,30 @@
...
+    $PathFilters = "--path '$VersionFilter/runtime/$OSFilter' --path '$VersionFilter/sdk/$OSFilter'"
+}
+else {
+    $PathFilters = "--path '$VersionFilter/$ImageTypeFilter/$OSFilter'"
 }
 
-$OptionalImageBuilderArgs += " --manifest $manifest"
+$OptionalImageBuilderArgs += " --manifest manifest.json"
 
 & ./eng/common/build-and-test.ps1 `
```
MichaelSimons: It looks like this would be running all of the tests not just the ones for desired image type/repo.

build-and-test.ps1
mthalman: I can rename it to `RepoFilter` but I can't make use of the `repo` option in ImageBuilder since that is limited to a single value.  So I'd still have to use the `path` option.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -147,25 +147,24 @@ private static string GetFilterRegexPattern(string filter)
...
+        [Trait("Category", "sdk")]
         [MemberData(nameof(GetVerifyRuntimeImagesData))]
         public void VerifyImagesWithWebApps(ImageDescriptor imageDescriptor)
         {
             VerifyFxImages(imageDescriptor, "webapp", "powershell -command \"dir ./bin/SimpleWebApplication.dll\"", false);
         }
 
         [Theory]
-        [Trait("Category", "all")]
         [Trait("Category", "ASPNET")]
```
MichaelSimons: Consider refactoring to make the casing align the the `runtime` and `sdk`.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -174,7 +173,6 @@ public void VerifyAspnetImagesWithApps(ImageDescriptor imageDescriptor)
         }
 
         [SkippableTheory("3.5")]
-        [Trait("Category", "all")]
         [Trait("Category", "WCF")]
```
MichaelSimons: Consider refactoring to make the casing align the the `runtime` and `sdk`.

build-and-test.ps1
```diff
@@ -1,22 +1,27 @@
 #!/usr/bin/env pwsh
 [cmdletbinding()]
 param(
-    [ValidateSet('*', 'runtime-sdk', 'aspnet', 'wcf')]
-    [string]$ImageTypeFilter = '*',
+    [ValidateSet('*', 'runtime', 'sdk', 'aspnet', 'wcf')]
```
MichaelSimons: It seems like '*' could be excluded.  The default could be an empty array which would indicate no filtering.  It would arguably make more sense because it wouldn't allow values like `runtime,*`.  Does `run-test` work with `runtime,*`?

build-and-test.ps1
```diff
@@ -1,22 +1,27 @@
...
+    [array]$RepoFilter = @('*'),
```
MichaelSimons: Would it be more valuable to declare this as a `string[]`?

tests/run-tests.ps1
```diff
@@ -47,5 +47,30 @@ else {
     $env:LOCAL_RUN = $null
 }
 
-& dotnet test --filter Category=$TestCategory -c Release --logger:trx $PSScriptRoot/Microsoft.DotNet.Framework.Docker.Tests/Microsoft.DotNet.Framework.Docker.Tests.csproj
+# Powershell treats each variable passed to the command line call as an individual argument.  But we're constructing the $testFilter variable
```
MichaelSimons: long comment lines

MichaelSimons: It feels like this logic would be simpler in a way to use the [Exec function](https://github.com/dotnet/docker-tools/blob/master/eng/common/Invoke-ImageBuilder.ps1#L45) we have in the common scripts.   The logic would build up a single Args string.  An added bonus is the Exec method would emit out the command that is being executed which would be helpful diagnostic info.

tests/run-tests.ps1
```diff
@@ -9,7 +9,7 @@ param(
     [string]$OSFilter,
     [string]$Registry,
     [string]$RepoPrefix,
-    [string]$TestCategory = "all",
+    [array]$TestCategories,
```
MichaelSimons: Shouldn't this be identical to the build-and-test script parameter? It should have the ValidateSet attribute.

tests/run-tests.ps1
```diff
@@ -9,11 +9,22 @@ param(
...
+    [ValidateSet('runtime', 'sdk', 'aspnet', 'wcf')]
+    [string[]]$TestCategories = @(),
     [switch]$IsLocalRun,
     [string]$ImageInfoPath
 )
 
+function Exec {
+    param ([string] $Cmd)
+
+    Log "Executing: '$Cmd'"
```
MichaelSimons: The `Log` function needs to be copied over as well.

PR #401, opened 2019-09-02
Updated ltsc2019 for newer version of Nuget CLI

r4nc0r:


mthalman:
@r4nc0r - Thank you for your contribution.  Due to the potential breaking changes between `4.4.1` and `5.1.0`, we're not going to take this change.  It could break consumers of this image.  Instead, you'll need to copy over a newer NuGet client in your own Dockerfile that is based on this image.

PR #408, opened 2019-09-10
Sept Patch Tuesday updates

mthalman:


MichaelSimons:
The repo readme updates are missing.

Issue #407, opened 2019-09-05
Enable xUnit parallelization in tests


mthalman:
The xUnit test framework has the capability to execute tests in parallel as long as they are in separate test classes.  Currently all the tests are in one class so they execute sequentially.  Experimentation has found that by executing the tests in parallel, we can improve test execution time by about 20%.

Executing the tests in parallel surfaces a few issues:
* A [Windows container bug](https://github.com/moby/moby/issues/37562) can occur when deleting a container.  Need to see whether this can be worked around by locking around stopping/removing a container.
* https://github.com/microsoft/dotnet-framework-docker/issues/400 occurs more frequently

MichaelSimons:
We should consider doing this for the .NET Core repo as well.  @mthalman, can you log an issue for that?

MichaelSimons:
This is currently blocked by #400

mthalman:
This issue is being close as won't fix due to an inability to workaround the [Windows container bug](https://github.com/moby/moby/issues/37562) without adversely affecting the performance and thereby losing the benefits hoped to be gained by the parallelization.

PR #410, opened 2019-09-11
Increase retries and bump up the sleep intervals 

HongGit:
To address the random failures, increase the retries to 60 and bump up the sleep intervals.

### DIFF
--- a/tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
+++ b/tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
@@ -321,7 +321,7 @@ namespace Microsoft.DotNet.Framework.Docker.Tests
 
         private void VerifyHttpResponseFromContainer(string containerName, string urlPath)
         {
-            var retries = 30;
+            var retries = 60;
 
             // Can't use localhost when running inside containers or Windows.
             var url = $"http://{_dockerHelper.GetContainerAddress(containerName)}" + urlPath;
@@ -331,7 +331,7 @@ namespace Microsoft.DotNet.Framework.Docker.Tests
                 while (retries > 0)
                 {
                     retries--;
-                    Thread.Sleep(TimeSpan.FromSeconds(2));
+                    Thread.Sleep(TimeSpan.FromSeconds(5));
                     try
                     {
                         using (HttpResponseMessage result = client.GetAsync(url).Result)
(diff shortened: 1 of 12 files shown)


tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -321,7 +321,7 @@ private static string GetManifestRegistry()
 
         private void VerifyHttpResponseFromContainer(string containerName, string urlPath)
         {
-            var retries = 30;
+            var retries = 60;
```
mthalman: This change can be reverted.  The original retry count of 30 works fine when the Dockerfile change is applied.

tests/Microsoft.DotNet.Framework.Docker.Tests/ImageTests.cs
```diff
@@ -331,7 +331,7 @@ private void VerifyHttpResponseFromContainer(string containerName, string urlPat
                 while (retries > 0)
                 {
                     retries--;
-                    Thread.Sleep(TimeSpan.FromSeconds(2));
+                    Thread.Sleep(TimeSpan.FromSeconds(5));
```
mthalman: This change can be reverted.  The original sleep count of 2 seconds works fine when the Dockerfile change is applied.

tests/Microsoft.DotNet.Framework.Docker.Tests/projects/aspnet-4.8/Dockerfile
```diff
@@ -2,6 +2,8 @@ ARG BASE_ASPNET_IMAGE
 
 FROM $BASE_ASPNET_IMAGE
 
+RUN Add-WindowsFeature Web-AppInit -IncludeAllSubFeature
```
mthalman: This does the trick.  Please apply this change to the rest of the ASP.NET and WCF test project Dockerfiles except for 3.5 where this command is returning an error code.  That's ok, though, because I don't see timeout issues for 3.5.

mthalman:
Actually, I'm unsure that these changes really do fix the issue.  I see that the build for this PR has failed even though it was using the Dockerfile with the change applied to it.  I'm still investigating this with my own testing.

mthalman:
@HongGit - That last commit just applied the same changes as the previous one but just to the rest of the Dockerfiles.  Those changes don't actually solve the problem since the https://github.com/microsoft/dotnet-framework-docker/pull/410/commits/21ee9669c7b4a20ed140a41cfe92ed7e3967c2db commit had still caused a test failure.

MichaelSimons:
The work to [enable test parallelization](https://github.com/microsoft/dotnet-framework-docker/issues/407) is what magnified this issue.  That work was cut due to some issue with Windows Containers.  Closing this as we haven't seen the tests fail recently.

Issue #327, opened 2019-07-18
Error : Can't find the valid AspnetMergePath


mrmonday:
Using the following:

[WebApplication1.zip](https://github.com/microsoft/dotnet-framework-docker/files/3406208/WebApplication1.zip)

```powershell
Expand-Archive .\WebApplication1.zip -DestinationPath .
cd .\WebApplication1\
docker build --no-cache .
```

Gives:

```
Sending build context to Docker daemon  20.48kB
Step 1/4 : FROM mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2019 AS build
 ---> c4fa4ec9c73e
Step 2/4 : WORKDIR /app
 ---> Running in b9df7e3f017d
Removing intermediate container b9df7e3f017d
 ---> 24e5cf8e9f78
Step 3/4 : COPY . .
 ---> 5e7fa420ccd0
Step 4/4 : RUN msbuild.exe /restore /property:DeployOnBuild=true /property:PublishProfile=FolderProfile WebApplication1.csproj
 ---> Running in 7e9c2e22a20c
Microsoft (R) Build Engine version 16.1.76+g14b0a930a7 for .NET Framework
Copyright (C) Microsoft Corporation. All rights reserved.

Build started 7/18/2019 12:06:30 PM.
Project "C:\app\WebApplication1.csproj" on node 1 (Restore target(s)).
Restore:
  Restoring packages for C:\app\WebApplication1.csproj...
    GET https://api.nuget.org/v3-flatcontainer/microsoft.codedom.providers.dotnetcompilerplatform/index.json
    OK https://api.nuget.org/v3-flatcontainer/microsoft.codedom.providers.dotnetcompilerplatform/index.json 133ms
    GET https://api.nuget.org/v3-flatcontainer/microsoft.codedom.providers.dotnetcompilerplatform/2.0.1/microsoft.codedom.providers.dotnetcompilerplatform.2.0.1.nupkg
    OK https://api.nuget.org/v3-flatcontainer/microsoft.codedom.providers.dotnetcompilerplatform/2.0.1/microsoft.codedom.providers.dotnetcompilerplatform.2.0.1.nupkg 15ms
  Installing Microsoft.CodeDom.Providers.DotNetCompilerPlatform 2.0.1.
  Committing restore...
  Generating MSBuild file C:\app\obj\WebApplication1.csproj.nuget.g.props.
[...]

Issue #323, opened 2019-07-15
error MSB3086: Task could not find "sgen.exe", seems load a wrong version


zhangxin511:
I am using `mcr.microsoft.com/dotnet/framework/sdk:4.8`, got following errors:
```
GenerateSerializationAssemblies target) -> 
  C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(3533,5): error MSB3086: Task could not find "sgen.exe" using the SdkToolsPath "" or the registry key "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\NETFXSDK\4.7.2\WinSDK-NetFx40Tools-x86". Make sure the SdkToolsPath is set and the tool exists in the correct processor specific location under the SdkToolsPath and that the Microsoft Windows SDK is installed 
```

But run following command it gives 
```
λ docker run --rm mcr.microsoft.com/dotnet/framework/sdk:4.8 where sgen.exe
C:\Program Files (x86)\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\sgen.exe
```

Seems the Microsoft.Common.CurrentVersion.targets load the SdkToolsPath  wrong?

mthalman:
Same sort of issue as #313

MichaelSimons:
This should be addressed with VS 16.3.  Leaving open to validate.

Issue #313, opened 2019-07-08
al.exe missing on sdk 4.8


RMarinaro:
Trying to build a project that relies on the build tools on the 4.8 images throws an error:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(3628,5): error MSB3086: Task could not find "al.exe" using the SdkToolsPath "" or the registry key "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Microsoft SDKs\NETFXSDK\4.7.2\WinSDK-NetFx40Tools-x86". Make sure the SdkToolsPath is set and the tool exists in the correct processor specific location under the SdkToolsPath and that the Microsoft Windows SDK is installed [C:\codebuild\tmp\output\src323577031\vendor-portals\BCS.VendorPortals\BCS.VendorPortals.Core\BCS.VendorPortals.Core.vbproj] 
Done Building Project "C:\codebuild\tmp\output\src323577031\vendor-portals\BCS.VendorPortals\BCS.VendorPortals.Core\BCS.VendorPortals.Core.vbproj" (default targets) -- FAILED. 
Done Building Project "C:\codebuild\tmp\output\src323577031\vendor-portals\BCS.VendorPortals\BCS.VendorPortals.WebApp\BCS.VendorPortals.WebApp.vbproj" (DeployOnServer target(s)) -- FAILED. 
```

It looks like it is trying to look for the SDK tools in the 4.7.2 directory instead of 4.8.  Reverting the SDK image to 4.7.2 fixed the build

I think this is related to https://github.com/microsoft/dotnet-framework-docker/pull/169

mthalman:
I believe this is related to #214, but in the context of 4.8 instead of 4.7.2.

MichaelSimons:
Assigning to @mthalman, to investigate.

Smenus:
I have my own version of the sdk image, because we needed to add a few extra things during the Build Tools install.
I've added this line to the end of the SDK install bit, and it does mean that the .exes from the 4.8 SDK are found by msbuild...
```powershell
[...]

PR #417, opened 2019-09-23
VS 16.3 updates

### LINKED ISSUE #276: Install .NET Fx 4.8 from VS build tools when it's available
Currently the 4.8 Dockerfiles are referencing a standalone MSI to install 4.8 since it's not a shipping component in VS 2019 currently.  When 4.8 is available with the VS installer, the Dockerfiles should be updated to make use of that instead.

mthalman:
* Updates the SDK images to the VS 16.3 versions of the Test Agent and Build Tools
* VS 16.3 now includes 4.8 so the 3.5 and 4.8 SDK images are updated to install 4.8 from Build Tools instead of from the blob storage install.

Fixes #276, #313, #323, #327

### DIFF
--- a/3.5/sdk/windowsservercore-1903/Dockerfile
+++ b/3.5/sdk/windowsservercore-1903/Dockerfile
@@ -11,17 +11,18 @@ RUN mkdir "%ProgramFiles%\NuGet" `
 # Install VS components
 RUN `
     # Install VS Test Agent
-    curl -fSLo vs_TestAgent.exe https://download.visualstudio.microsoft.com/download/pr/73633735-85b4-4b25-9e30-10eedcc8bbb7/ad86fa8eda143cdca5f4944e30135c79/vs_testagent.exe `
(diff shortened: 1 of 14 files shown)


3.5/sdk/windowsservercore-1903/Dockerfile
```diff
@@ -57,6 +50,19 @@ RUN `
     `
     # Workaround VS installer/ngen issues
     && \Windows\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\TestAgent\Common7\IDE\VSWebLauncher.exe" `
+    && \Windows\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin\Roslyn\csc.exe.config" `
```
rainersigwald: I'm confused by this. Can you elaborate on why it's useful?

My understanding is that since this isn't a managed assembly, it can't ever be ngened, so uninstalling it will always fail, short-circuiting the rest of this statement.

mthalman: It's a workaround for https://github.com/dotnet/roslyn/issues/38296

mthalman:
Due to an issue with installing the VS Build Tools for the 4.7.1 SDK image, that image is not being updated with these changes.

PR #422, opened 2019-09-27
Adjusting build timeouts after upgrading build hardware

### LINKED ISSUE #414: Adjust build timeout values now that build times have improved
Due to machine infrastructure changes, build times have improved.  The timeout values should be adjusted based on the new build times to ensure we don't waste a ton of time on edge cases where the build isn't working correctly.

MichaelSimons:
Fixes #414

### DIFF
--- a/eng/pipelines/dotnet-framework.yml
+++ b/eng/pipelines/dotnet-framework.yml
@@ -11,8 +11,9 @@ stages:
 - template: ../common/templates/stages/build-test-publish-repo.yml
   parameters:
     ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-      windowsAmdBuildJobTimeout: 150
+      windowsAmdBuildJobTimeout: 240
+      windowsAmdTestJobTimeout: 90
     ${{ if eq(variables['System.TeamProject'], 'public') }}:
       buildMatrixType: platformVersionedOs
       customBuildLegGrouping: pr-build
-      windowsAmdBuildJobTimeout: 300
+      windowsAmdBuildJobTimeout: 330
--- a/eng/pipelines/dotnet-framework.yml
+++ b/eng/pipelines/dotnet-framework.yml
@@ -11,9 +11,8 @@ stages:
 - template: ../common/templates/stages/build-test-publish-repo.yml
   parameters:
     ${{ if eq(variables['System.TeamProject'], 'internal') }}:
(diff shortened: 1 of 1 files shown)


eng/pipelines/dotnet-framework.yml
```diff
@@ -11,9 +11,8 @@ stages:
...
   parameters:
     ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-      windowsAmdBuildJobTimeout: 300
-      windowsAmdTestJobTimeout: 120
+      windowsAmdBuildJobTimeout: 150
     ${{ if eq(variables['System.TeamProject'], 'public') }}:
       buildMatrixType: platformVersionedOs
       customBuildLegGrouping: pr-build
-      windowsAmdBuildJobTimeout: 420
+      windowsAmdBuildJobTimeout: 300
```
mthalman: Considering that the build for this PR had a leg that took 4:50, I think this should at least be changed to 330.

eng/pipelines/dotnet-framework.yml
```diff
@@ -11,9 +11,8 @@ stages:
 - template: ../common/templates/stages/build-test-publish-repo.yml
   parameters:
     ${{ if eq(variables['System.TeamProject'], 'internal') }}:
-      windowsAmdBuildJobTimeout: 300
-      windowsAmdTestJobTimeout: 120
+      windowsAmdBuildJobTimeout: 150
```
mthalman: This is much too small.  There was a leg on Patch Tuesday that took 190 minutes.  I really, really don't want to see timeouts for official releases so we need to have a sufficient buffer, especially considering the extra time it takes to pull down the new Windows images.  I'd be more comfortable with this set to 240 and we can adjust as we get more data.  There can be a lot of variability in build times and I don't want to be too aggressive when setting timeouts.  Also, some test jobs take about 54 minutes which is too close for comfort compared to the default 60 minutes.  That should be set to 90 minutes.

eng/pipelines/dotnet-framework.yml
MichaelSimons: Changed as suggested

eng/pipelines/dotnet-framework.yml
MichaelSimons: good feedback - changed as suggested.

PR #435, opened 2019-10-24
The ` is not used as a softline seperator!, but the \ is!

gesture1968:
Please change all the ` into \ so the dockerfile is working correctly

### DIFF
--- a/4.8/runtime/windowsservercore-ltsc2016/Dockerfile
+++ b/4.8/runtime/windowsservercore-ltsc2016/Dockerfile
@@ -1,32 +1,33 @@
 # escape=`
+# soft newline =\
 
 FROM mcr.microsoft.com/windows/servercore:ltsc2016
 
 # Install .NET 4.8
-RUN powershell -Command `
-        $ProgressPreference = 'SilentlyContinue'; `
-        Invoke-WebRequest `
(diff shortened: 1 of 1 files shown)


mthalman:
The Dockefile does work correctly because of the escape directive at the top:

```
# escape=`
```

This is documented here: https://docs.docker.com/engine/reference/builder/#escape.  This is used because of the following quote from that link:

> Setting the escape character to ` is especially useful on Windows, where \ is the directory path separator. ` is consistent with Windows PowerShell.

Were you running into issues with building this Dockerfile?

4.8/runtime/windowsservercore-ltsc2016/Dockerfile
```diff
@@ -1,32 +1,33 @@
 # escape=`
+# soft newline =\
```
MichaelSimons: @gesture1968, Can you point me to the documentation for this?  I have not seen this syntax before.

gesture1968:
Hello Matt,

Yes, I had problems and indeed I did mistake the '# escape' for comment 🙁, so I left it out. That also explains why I had to change the path \ to \\ in my existing dockerfile. Thanks for making that clear, sorry for the inconvenience.

Greetings
Gesture

________________________________

PR #446, opened 2019-11-12
Merge changes for 11B release

mthalman:


MichaelSimons:
Friendly reminder to not squash this when merging.  I think we should preserve the commit history for this one.

PR #457, opened 2019-11-20
Corrected references to WCF to use Foundation instead of Framework.

### LINKED ISSUE #449: Fix WCF terminology
There are many cases in readme files where WCF is referred to as "Windows Communication **Framework**" when it should be referred to as "Windows Communication **Foundation**".

StephenBonikowsky:
I think I got them all.
Fixes #449


StephenBonikowsky:
Created microsoft/mcr#798

PR #460, opened 2019-11-25
CLI tool to update dependencies

### LINKED ISSUE #439: update-dependencies tool to automate changes for each release
To help eliminate human error when preparing changes for releases, we should automate the process of updating the usual set of files.  Examples of updates it would make include:
* timestamp values in manifest.json
* tag values in readme
* MSU patch files
* NuGet CLI version
* VS build tools URLs

This would work similarly to the [update-dependencies](https://github.com/dotnet/dotnet-docker/tree/master/eng/update-dependencies) tool in dotnet-docker.

mthalman:
Provides a tool for automatically updating the following aspects of .NET Fx Dockerfiles:
* LCU download URLs
* Version numbers in manifest.json
* VS tools download URLs
* NuGet client version
* Readme tag updates

For the LCU, VS Tools, and NuGet client version, it uses JSON files in the `eng` folder to define the master values to read from and propagate to the Dockerfiles.

Fixes #439


eng/update-dependencies/DependencyUpdater.cs
```diff
@@ -0,0 +1,303 @@
...
+                    new CustomFileRegexUpdater(nuGetVersion.Version, file.BuildInfoName)
+                    {
+                        Path = file.Path,
+                        VersionGroupName = NuGetVersionGroupName,
+                        Regex = new Regex($"ENV NUGET_VERSION (?<{NuGetVersionGroupName}>{nuGetVersion.Filter.Replace(".", @"\.").Replace("*", $@"\d*")})")
+                    }
+                ));
+        }
+
+        private static IEnumerable<DockerfileInfo> GetDockerfiles()
```
MichaelSimons: Is it worthwhile to cache the results given the multiple times it is called?

eng/nuget-client-versions.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "filter": "4.4.*",
```
MichaelSimons: I was expecting a pattern more like the lcu-config where the Fx versions were specified to determine the behavior rather than a wild card pattern.  I thought this because of my experiences with .NET Core.  When new .NET versions come out, you can easily copy the previous Dockerfiles and use the updater tool to modify the copied Dockerfiles.  With this pattern both the current and new Dockerfiles would get updated.  In reality this mater here since 4.8 is the last .NET FX version.

I think the `filter` property could be given a more descriptive name to make it more meaningful.  Maybe MatchVersionFilter?

eng/nuget-client-versions.json
```diff
@@ -0,0 +1,10 @@
+[
```
MichaelSimons: I think it would be nice if the three json files had a similar naming convention.  I understand nuget-config.json is confusing.  `lcu-info.json`, `nuget-client.json`, `vs-info.json`

eng/update-dependencies/DependencyUpdater.cs
```diff
@@ -0,0 +1,303 @@
...
+                return null;
+            }
+
+            LcuConfig? lcuConfig = lcuConfigs
+                .FirstOrDefault(config => config.OsVersion == dockerfile.OsVersion &&
+                    config.RuntimeVersions.Any(runtime => runtime == dockerfile.FrameworkVersion));
+
+            if (lcuConfig != null)
+            {
+                return (lcuConfig, dockerfile.BuildInfoName);
```
MichaelSimons: LCU only applies to the Runtime Dockerfiles.  Why does that need to be returned versus hard coding it in CreateLcuUpdaters?  Avoiding the anonymous type would help simplify the code.

eng/nuget-client-versions.json
mthalman: Except it's not going to vary by framework version; it'll vary by OS version.  Would you be ok with something like this:

```JSON
[
  {
    "nugetClientVersion": "4.4.3",
    "osVersions": [
      "windowsservercore-1903",
      "windowsservercore-ltsc2016",
      "windowsservercore-ltsc2019"
    ]
  },
  {
    "nugetClientVersion": "5.3.1",
    "osVersions": [
      "windowsservercore-1909"
    ]
  }
]
```

MichaelSimons: Good point on the OS version.  Yes I like this format.

eng/update-dependencies/DependencyUpdater.cs
```diff
@@ -0,0 +1,303 @@
...
+        private IEnumerable<IDependencyUpdater> CreateVsUpdaters()
+        {
+            VsConfig vsConfig = JsonConvert.DeserializeObject<VsConfig>(
+                File.ReadAllText(this.options.VsConfigPath));
+
+            const string TestAgentGroupName = "TestAgent";
+            const string BuildToolsGroupName = "BuildTools";
+            const string WebTargetsGroupName = "WebTargets";
+
+            return GetDockerfiles()
```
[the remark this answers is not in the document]
mthalman: Added a `Where` method.  I no longer have a `GetDockerfiles` method to pass in a parameter since it's now cached.

eng/update-dependencies/DependencyUpdater.cs
```diff
@@ -0,0 +1,312 @@
...
+                .FirstOrDefault(config => config.OsVersion == dockerfile.OsVersion &&
+                    config.RuntimeVersions.Any(runtime => runtime == dockerfile.FrameworkVersion));
+        }
+
+        private IEnumerable<IDependencyUpdater> CreateNuGetUpdaters()
+        {
+            NuGetInfo[] nuGetVersions = JsonConvert.DeserializeObject<NuGetInfo[]>(
+                File.ReadAllText(this.options.NuGetInfoPath));
+
+            // Verify a given OS version is not specified more than once
```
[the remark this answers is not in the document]
mthalman: On second thought, it's probably unnecessary code.  There really isn't going to be manipulation of the content of an existing osVersions array other than to remove EOL versions.  The impact of having duplicates would just be that the last regex updater for that given Dockerfile wins.

Issue #86, opened 2018-01-31
Create more extensive automated tests for SDK images


MichaelSimons:
Some of the secondary components that are included in the build images are not covered by the automated tests that run as part of CI.  Examples include VS Test Agent, Nuget CLI, PATH ENV contains the expected paths, etc.

mthalman:
Also verify:
* WebDeploy installation
* ClickOnce installation
* PowerShell should be set as the default shell

PR #476, opened 2019-12-16
Add test to verify vswhere tool works in SDK images

mthalman:
Related to #86


mthalman:
I needed to update the test project to 3.1 so it can reference the Models project.  The Models project uses `<nullable>enable</nullable>` so it requires C# 8.0 which .NET 2.1 didn't support.

PR #485, opened 2020-01-08
Remove escaping backslashes.

### LINKED ISSUE #474: ROSLYN_COMPILER_LOCATION environment variable is not set correctly in the ASP.NET images
```
> docker run -it --rm --entrypoint cmd mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-1909 cmd /c "set"
...
ROSLYN_COMPILER_LOCATION=c:\\RoslynCompilers\\tools
...
```

See https://github.com/microsoft/dotnet-framework-docker/blob/master/4.8/aspnet/windowsservercore-1909/Dockerfile#L25

The back slashes are needlessly escaped.

StephenBonikowsky:
* They are not needed.
Fixes #474


StephenBonikowsky:
I found a static string in test code that contained the escaped backslashes. Updated.

PR #487, opened 2020-01-09
Refactor build script to follow dotnet-docker build script pattern

mthalman:
Updates the `build-and-test.ps1` script to follow the same pattern used by https://github.com/dotnet/dotnet-docker/blob/master/build-and-test.ps1.  This defines `-BuildOnly` and `-TestOnly` parameters.  This is in preparation for the changes to remove testing logic from the common build script (https://github.com/dotnet/docker-tools/issues/364).

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -27,7 +27,7 @@ else {
 
 if ($RepoFilter.Count -eq 0) {
     $PathFilters = $null
-    $testCategories = @("runtime", "sdk", "aspnet", "wcf")
+    $testCategories = @()
 }
 else {
     $PathFilters = ""
@@ -36,8 +36,6 @@ else {
     }
     $testCategories = $RepoFilter
 }
-
-$OptionalImageBuilderArgs += " --manifest manifest.json"
 
 if ($build) {
     & ./eng/common/build-and-test.ps1 `
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -1,36 +1,56 @@
 #!/usr/bin/env pwsh
-[cmdletbinding()]
+[cmdletbinding(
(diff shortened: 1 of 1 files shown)


build-and-test.ps1
```diff
@@ -1,36 +1,56 @@
...
         $PathFilters += " --path '$VersionFilter/$_/$OSFilter'"
     }
-    # Convert the array to a comma-delimited string of the repos with each repo value in quotes
-    # (e.g. "runtime", "sdk")
-    $repoList = ($RepoFilter | foreach { "`"$_`"" }) -join ", "
-    $optionalTestArgs = "-TestCategories @($repoList)"
+    $testCategories = $RepoFilter
 }
 
 $OptionalImageBuilderArgs += " --manifest manifest.json"
```
MichaelSimons: Thoughts on removing this?  Is it necessary?  Doesn't it default to this?  It isn't specified in the core repo.

build-and-test.ps1
```diff
@@ -1,36 +1,56 @@
...
+}
+else {
+    $build = $BuildOnly
+    $test = $TestOnly
+}
+
 if ($RepoFilter.Count -eq 0) {
     $PathFilters = $null
-    $optionalTestArgs = ""
+    $testCategories = @("runtime", "sdk", "aspnet", "wcf")
```
MichaelSimons: Couldn't this be left empty?  It would eliminate the maintenance of having the full list declared multiple times within the script.

build-and-test.ps1
mthalman: Makes sense.  I think this was leftover from when we had separate manifest for aspnet/wcf.

PR #506, opened 2020-02-05
Add issue templates

mthalman:
Defines two markdown files to be used as templates for creating issues for a release. These are stored in a `releases` subfolder since they are not intended to be available to select when creating a new issue.

Also includes a bug report issue template for the repo.

Related to dotnet/docker-tools#383

### DIFF
--- a/.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
+++ b/.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
@@ -4,18 +4,20 @@ VS Version: &lt;version&gt;
 
 ## Tasks
 
+_There is overlap between the tasks here and those for a [Patch Tuesday release](patch-tuesday-release.md). In the case where a VS release is on the same day as Patch Tuesday, the tasks should be combined into one workflow. Tasks specific to only a VS release have been **bolded**._
+
 1. - [ ] Merge any pending PRs or commits from `dev` branch:
       - [ ] &lt;placeholder link&gt;
-1. - [ ] Update the `vsVersion` in [vs-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/vs-info.json) to be the new VS version being released
 1. - [ ] Look up the latest [NuGet CLI versions](https://www.nuget.org/downloads) and update the [nuget-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/nuget-info.json) file with the latest patch of each major/minor version listed in the file
-1. - [ ] Wait for VS to be released
-1. - [ ] Install or update the new version of VS on your local machine
-1. - [ ] Create a zip file named `MSBuild.Microsoft.VisualStudio.Web.targets.<YYYY>.<MM>.zip` that contains the `MSBuild\Microsoft\VisualStudio\v16.0\Web` and `MSBuild\Microsoft\VisualStudio\v16.0\WebApplications` folders from the VS installation location
-1. - [ ] Upload the zip file to https://dotnetbinaries.blob.core.windows.net/dockerassets
-1. - [ ] Update the `webTargetsUrl` in [vs-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/vs-info.json) to reference the new zip file name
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
```diff
@@ -0,0 +1,30 @@
...
+1. - [ ] Run the `update-dependencies` tool to update all the necessary files:
+      - [ ] `dotnet run --project .\eng\update-dependencies\update-dependencies.csproj --datestamp-all <YYYYMMDD>`
+1. - [ ] Inspect generated changes for correctness
+1. - [ ] Commit generated changes
+1. - [ ] Create PR
+1. - [ ] Get PR signoff
+1. - [ ] Merge PR
+1. - [ ] Wait for changes to be mirrored to internal [dotnet-framework-docker repo](https://dev.azure.com/dnceng/internal/_git/dotnet-dotnet-framework-docker) (internal MSFT link)
+1. - [ ] Queue build of [dotnet-framework-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=372) (internal MSFT link)
```
MichaelSimons: This build should only be building the SDK images.  Additionally, the Samples do not need to be rebuilt because the runtime images haven't changed.

.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
```diff
@@ -0,0 +1,30 @@
+# VS Tools Release
+
+VS Version: &lt;version&gt;
+
+## Tasks
```
MichaelSimons: Is it worthwhile calling out that VS releases should be combined with the PT release process when they happen on the same day?

mthalman:
> Thoughts on adding a template for onboarding a new OS version? Doesn't have to be included with this PR but feels like something that should happen as part of preparing for the 20H1 release.

I'm unsure of whether that should be documented in GitHub.  Certainly it's not really specific to .NET Fx and would apply to .NET Core.  Would we similarly provide documentation for removal of OS support?

MichaelSimons:
> >Thoughts on adding a template for onboarding a new OS version? Doesn't have to be included with this PR but feels like something that should happen as part of preparing for the 20H1 release.

> I'm unsure of whether that should be documented in GitHub. Certainly it's not really specific to .NET Fx and would apply to .NET Core. Would we similarly provide documentation for removal of OS support?

To me it fits the theme of https://github.com/dotnet/docker-tools/issues/383 as it is just another type of release.  I agree it is not specific to .NET Fx but there are a few actions that are specific to the .NET Fx images.  Capturing the steps to remove an OS should be part of the [larger epic](https://github.com/dotnet/docker-tools/issues/382).

.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
mthalman: We don't have support for only building the SDK images because of complications with the `REPO` arg getting set to the ACR location but needing to override that.  But we should be ensuring that only the SDK timestamps get changed.  I can make that adjustment.

.github/ISSUE_TEMPLATE/releases/vs-tools-release.md
mthalman: Fixed

mthalman:
> > > Thoughts on adding a template for onboarding a new OS version? Doesn't have to be included with this PR but feels like something that should happen as part of preparing for the 20H1 release.
> 
> > I'm unsure of whether that should be documented in GitHub. Certainly it's not really specific to .NET Fx and would apply to .NET Core. Would we similarly provide documentation for removal of OS support?
> 
> To me it fits the theme of [dotnet/docker-tools#383](https://github.com/dotnet/docker-tools/issues/383) as it is just another type of release. I agree it is not specific to .NET Fx but there are a few actions that are specific to the .NET Fx images. Capturing the steps to remove an OS should be part of the [larger epic](https://github.com/dotnet/docker-tools/issues/382).

I've updated https://github.com/dotnet/docker-tools/issues/383 to capture this.  Will be done in separate PR.

Issue #513, opened 2020-02-07
Patch Tuesday Release - 2B


mthalman:
# Patch Tuesday Release - 2B

## Tasks

1. - [x] Merge any pending PRs or commits from `dev` branch:
      - [x] &lt;placeholder link&gt;
1. - [x] Look up the latest [NuGet CLI versions](https://www.nuget.org/downloads) and update the [nuget-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/nuget-info.json) file with the latest patch of each major/minor version listed in the file.
1. - [x] Wait for latest cumulative updates (LCUs) to be released (typically at 10 AM PST on Patch Tuesday).
1. - [x] Gather list of KB numbers for the .NET Framework updates from the .NET Release team.
1. - [x] Look up the download URL for each of the KB numbers in [Microsoft Update Catalog](https://www.catalog.update.microsoft.com/) and input them into the [lcu-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/lcu-info.json) file.
1. - [x] Run the `update-dependencies` tool to update all the necessary files:
      - [x] `dotnet run --project .\eng\update-dependencies --datestamp-all <YYYYMMDD>`
1. - [x] Inspect generated changes for correctness
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR
1. - [x] Wait for changes to be mirrored to internal [dotnet-framework-docker repo](https://dev.azure.com/dnceng/internal/_git/Microsoft-dotnet-framework-docker) (internal MSFT link)
1. - [x] Wait for the following Windows images to have been updated as part of the Windows Patch Tuesday release process (this begins at 10 AM PST on Patch Tuesday):
      - [x] `mcr.microsoft.com/windows/servercore:1903`
      - [x] `mcr.microsoft.com/windows/servercore:1909`
      - [x] `mcr.microsoft.com/windows/servercore:ltsc2016`
      - [x] `mcr.microsoft.com/windows/servercore:ltsc2019`
1. - [x] Queue build of [dotnet-framework-docker pipeline](https://dev.azure.com/dnceng/internal/_build?definitionId=372) (internal MSFT link)
1. - [x] Confirm images have been ingested by MCR
1. - [x] Confirm READMEs have been updated in Docker Hub for [microsoft-dotnet-framework](https://hub.docker.com/_/microsoft-dotnet-framework)
[...]

PR #516, opened 2020-02-11
Updates for 2B patches

mthalman:
Related to https://github.com/microsoft/dotnet-framework-docker/issues/513


mquekel:
Hello, I see that the 11feb Windows update was added. But it looks like Docker Hub does not use this version yet. https://hub.docker.com/_/microsoft-dotnet-framework-aspnet/?tab=description

When will the new images get pushed to Docker Hub?

Issue #533, opened 2020-03-03
Patch Tuesday Release - 3B


mthalman:
# Patch Tuesday Release

## Tasks

1. - [x] Merge any pending PRs or commits from `dev` branch:
      - [x] &lt;placeholder link&gt;
1. - [x] Look up the latest [NuGet CLI versions](https://www.nuget.org/downloads) and update the [nuget-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/nuget-info.json) file with the latest patch of each major/minor version listed in the file.
1. - [x] Wait for latest cumulative updates (LCUs) to be released (typically at 10 AM PST on Patch Tuesday).
1. - [x] Gather list of KB numbers for the .NET Framework updates from the .NET Release team.
1. - [x] Look up the download URL for each of the KB numbers in [Microsoft Update Catalog](https://www.catalog.update.microsoft.com/) and input them into the [lcu-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/lcu-info.json) file. If this is the first Patch Tuesday after the release of a new Windows version, you'll need to do the following extra steps for that version:
      - [ ] Add a new entry to the [lcu-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/lcu-info.json) file to associate a URL for the new Windows version and `3.5` runtime version.
      - [ ] Update the `3.5/runtime` Dockerfile for the new Windows version so that the patch is applied (this can be copied from another `3.5/runtime` Dockerfile in a section labeled `Apply latest patch`)
1. - [x] Run the `update-dependencies` tool to update all the necessary files:
      - [x] `dotnet run --project .\eng\update-dependencies --datestamp-all <YYYYMMDD>`
1. - [x] Inspect generated changes for correctness
1. - [x] Commit generated changes
1. - [x] Create PR
1. - [x] Get PR signoff
1. - [x] Merge PR
1. - [x] Wait for changes to be mirrored to internal [dotnet-framework-docker repo](https://dev.azure.com/dnceng/internal/_git/Microsoft-dotnet-framework-docker) (internal MSFT link)
[...]

PR #538, opened 2020-03-10
Updates for 3B Patch Tuesday release

mthalman:
Related to https://github.com/microsoft/dotnet-framework-docker/issues/533

### DIFF
--- a/3.5/runtime/windowsservercore-ltsc2016/Dockerfile
+++ b/3.5/runtime/windowsservercore-ltsc2016/Dockerfile
@@ -21,12 +21,12 @@ RUN powershell -Command `
         $ProgressPreference = 'SilentlyContinue'; `
         Invoke-WebRequest `
             -UseBasicParsing `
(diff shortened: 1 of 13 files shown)


sanguedemonstro:
Hey guys @mthalman @MichaelSimons,
Apparently after commit [26597e4 ](https://github.com/microsoft/dotnet-framework-docker/commit/26597e42d157cc1e09d1e0dc8f23c32e6c3d1467) `4.7.2-runtime-windowsservercore-ltsc2019` image just grew from 4.2GB to 6.6GB.

Is that right?

mthalman:
@sanguedemonstro - I'm not seeing the same size for the original value as you provided.  The tag for the image before this commit was made is `mcr.microsoft.com/dotnet/framework/runtime:4.7.2-20200211-windowsservercore-ltsc2019`.  Its size is 5.1 GB.  So the change was from 5.1 GB to 6.6 GB.  

The installation of the patch was necessary because the base Windows image was not carrying a patched version of .NET Framework 4.7.2.

Issue #534, opened 2020-03-03
VS Tools Release - 16.5


mthalman:
# VS Tools Release

VS Version: 16.5

## Tasks

_There is overlap between the tasks here and those for a [Patch Tuesday release](patch-tuesday-release.md). In the case where a VS release is on the same day as Patch Tuesday, the tasks should be combined into one workflow. Tasks specific to only a VS release have been **bolded**._

1. - [x] Merge any pending PRs or commits from `dev` branch:
      - [ ] &lt;placeholder link&gt;
1. - [x] Look up the latest [NuGet CLI versions](https://www.nuget.org/downloads) and update the [nuget-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/nuget-info.json) file with the latest patch of each major/minor version listed in the file
1. - [x] **Update the `vsVersion` in [vs-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/vs-info.json) to be the new VS version being released**
1. - [x] **Wait for VS to be released**
1. - [x] **Install or update the new version of VS on your local machine**
1. - [x] **Create a zip file named `MSBuild.Microsoft.VisualStudio.Web.targets.<YYYY>.<MM>.zip` that contains the `MSBuild\Microsoft\VisualStudio\v16.0\Web` and `MSBuild\Microsoft\VisualStudio\v16.0\WebApplications` folders from the VS installation location**
1. - [x] **Upload the zip file to https://dotnetbinaries.blob.core.windows.net/dockerassets**
1. - [x] **Update the `webTargetsUrl` in [vs-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/vs-info.json) to reference the new zip file name**
1. - [x] **Gather the Build Tools and Test Agent download URLs by navigating to https://visualstudio.microsoft.com/downloads. Download links can be found under the `Tools for Visual Studio` section. Prior to clicking the download button, open a network profiler like your browser's dev tools or Fiddler so that you can capture the URL that gets downloaded. For `Build Tools for Visual Studio` look for a file named `vs_BuildTools.exe` being downloaded. For `Agents for Visual Studio`, ensure the radio button is defaulted to `Agent` and look for a file named `vs_TestAgent.exe`. Copy those two URLs and paste them into [vs-info.json](https://github.com/microsoft/dotnet-framework-docker/blob/master/eng/vs-info.json).**
[...]

PR #543, opened 2020-03-16
Reference VS 16.5 tools

mthalman:
Updates the SDK Dockerfiles to reference the newly released tools in VS 16.5.

Related to https://github.com/microsoft/dotnet-framework-docker/issues/534

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,7 +3,7 @@
   "registry": "mcr.microsoft.com",
   "variables": {
     "RuntimeReleaseDateStamp": "20200211",
-    "SdkReleaseDateStamp": "20200211",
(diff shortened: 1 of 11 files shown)


manifest.json
```diff
@@ -3,7 +3,7 @@
   "registry": "mcr.microsoft.com",
   "variables": {
     "RuntimeReleaseDateStamp": "20200211",
-    "SdkReleaseDateStamp": "20200211",
+    "SdkReleaseDateStamp": "20200316",
```
MichaelSimons: Shouldn't the `3B-*` variable references should be replaced with this variable w/this release?

mthalman: Good catch

PR #549, opened 2020-03-26
User/mbrown555/announce

mbrown555:


README.aspnet.md
```diff
@@ -1,6 +1,6 @@
 ## Important: Client Firewall Rules Update to Microsoft Container Registry (MCR)
 
-To provide a consistent FQDNs, on **March 3, 2020** the data endpoint will be changing from *.cdn.mscr.io to *.data.microsoft.com.
+To provide a consistent FQDNs, the data endpoint will be changing from *.cdn.mscr.io to *.data.mcr.microsoft.com
```
MichaelSimons: When is this change going to be made?

mbrown555: I was just going to message you because I didn't mean to merge this change, I thought I was just merging into my fork

mbrown555: I would've waited for you to approve. I mostly use branches in the OS repo, not very experienced with forks and I hit some issues. In any case, lmk if the change looks okay.

mbrown555: Change is supposed to happen in June but since this can change again I left it unspecified and just point to the document https://github.com/microsoft/containerregistry/blob/master/client-firewall-rules.md since that will be kept up to date.

PR #594, opened 2020-06-09
Updates for 6B release

mthalman:
Related to #593


mthalman:
I realized I updated tags for everything instead of just 2016.  I'll fix that.

mthalman:
Still need to update the readmes

PR #600, opened 2020-06-18
Make use of TLS 1.2 for downloads

### LINKED ISSUE #598: SDK for Server 2016 build fails downloading NuGet client
When attempting to build either of the 3.5 or 4.8 SDK Dockerfiles for Server 2016, it results in an error message when downloading the NuGet client: `The request was aborted: Could not create SSL/TLS secure channel`.  The affected line is here: https://github.com/microsoft/dotnet-framework-docker/blob/926f1556eab45d522969a0a07dc68c524cd6a48b/4.8/sdk/windowsservercore-ltsc2016/Dockerfile#L11

The error occurs because the default security protocol for Server 2016 is SSL3 or TLS 1.0 and nuget.org must have made a recent change to require TLS 1.2.

This issue can be resolved by setting the security protocol to TLS 1.2 prior to executing `Invoke-WebRequest`:

```powershell
[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```

mthalman:
Even though the build is broken due to the TLS requirement of dist.nuget.org, I felt it was appropriate to update _all_ cases where the Server 2016 Dockerfiles download files.  This ensures that any potential future TLS changes from those domains won't break the build.

Fixes #598

### DIFF
--- a/src/aspnet/4.7/windowsservercore-ltsc2016/Dockerfile
+++ b/src/aspnet/4.7/windowsservercore-ltsc2016/Dockerfile
@@ -9,6 +9,7 @@ RUN Add-WindowsFeature Web-Server; `
     Add-WindowsFeature NET-Framework-45-ASPNET; `
     Add-WindowsFeature Web-Asp-Net45; `
     Remove-Item -Recurse C:\inetpub\wwwroot\*; `
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; `
     Invoke-WebRequest -Uri https://dotnetbinaries.blob.core.windows.net/servicemonitor/2.0.1.10/ServiceMonitor.exe -OutFile C:\ServiceMonitor.exe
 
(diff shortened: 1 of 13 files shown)


src/aspnet/4.7/windowsservercore-ltsc2016/Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN Add-WindowsFeature Web-Server; `
     Add-WindowsFeature NET-Framework-45-ASPNET; `
     Add-WindowsFeature Web-Asp-Net45; `
     Remove-Item -Recurse C:\inetpub\wwwroot\*; `
+    [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; `
```
v-chayan, line 12: Should also add the SecurityProtocal setting in below "Install Roslyn compilers and ngen binaries" section

PR #606, opened 2020-07-09
Set 4.8-3.5 shared tag as undocumented

### LINKED ISSUE #604: Manifest for 4.8-3.5 tag includes versions of Windows that don't have 4.8 installed
The changes in https://github.com/microsoft/dotnet-framework-docker/commit/19559e06a9703dfde0a276589d353a9834787d78 added the `4.8-3.5` tag but it gets applied to all Windows versions.  This means you could pull that tag on a Windows Server 2016 machine and it will successfully pull down the 3.5 image.  But the 3.5 image on Server 2016 doesn't have 4.8 installed in it.  The `4.8-3.5` tag should really only apply to 1903 and later versions which do have 4.8 installed on them.

mthalman:
Fixes #604


MichaelSimons:
@mthalman, can you please update #604 to capture the discussion around why we are not going to support the `4.8-3.5` tag.

PR #613, opened 2020-07-14
Updates for 7B release

mthalman:
Related to https://github.com/microsoft/dotnet-framework-docker/issues/610


mthalman:
I forgot that we need to add `ngen update` to the the 4.8 runtime now that the patch is being applied.  Otherwise the small set of assemblies that are supposed to be ngen'd show up as queued.

PR #615, opened 2020-07-17
remove firewall notice add standard license link

mbrown555:
Need to remove the firewall notice and add a link to a standard license notice to all readmes

### DIFF
--- a/README.aspnet.md
+++ b/README.aspnet.md
@@ -1,10 +1,3 @@
-## Important: Client Firewall Rules Update to Microsoft Container Registry (MCR)
-
-To provide a consistent FQDNs, the data endpoint will be changing from *.cdn.mscr.io to *.data.mcr.microsoft.com
-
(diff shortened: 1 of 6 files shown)


mbrown555:
I've removed the extra line. This link was cleared by CELA. We'd like to check merge this as-is for now if it's okay with you, all the other Docker Hub sites have the same thing for now. In order to change it we need to get clearance for a change. I've added Toddy as a reviewer, he can comment further.

mbrown555:
Please review https://github.com/dotnet/dotnet-docker/pull/2097 as well, I'm not able to add reviewers to that one.

toddysm:
@mthalman aka.ms link was communicated to @OSSAttorney awhile back and we didn't hear any concerns. We have already updated all other pages except the .NET ones with that link and we will need to redo all this manual work. This semester we hope to have a better way to manage this content and we will update the link with your suggestion if required.
Besides that, the link is not very visible (Edge doesn't even show it in the status bar) and the redirect is done hidden from the user. I believe, it should be OK for now.

PR #637, opened 2020-09-11
Add readme templates

MichaelSimons:
Related to #https://github.com/dotnet/dotnet-docker/issues/1931

* Created a single readme template to generate all of the readmes from
* Created unit tests to ensure the readmes and templates are always in sync
* Fixed a few inconsistencies across the various readmes


tests/Microsoft.DotNet.Framework.Docker.Tests/GeneratedArtifactTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                generateTagsDocumentationScript,
+                $"The Readmes are out of sync with the templates.  Update the Readmes by running `{generateTagsDocumentationScript}`.");
+        }
+
+        private void ValidateGeneratedArtifacts(string generateScriptPath, string errorMessage)
+        {
+            string powershellArgs = $"-File {generateScriptPath} -Validate";
+            (Process Process, string StdOut, string StdErr) executeResult;
+
+            // Support both execution within Windows 10, Nano Server and Linux environments.
```
mthalman, line 37: Doesn't need to account for Nano Server and Linux in this repo.

mthalman (approved):

tests/Microsoft.DotNet.Framework.Docker.Tests/GeneratedArtifactTests.cs
MichaelSimons: It does actually need to support Linux.  The pre-build validation is run on a Linux build agent for PR validation and official builds.  When ltsc2016 no longer needs to be supported then we will be able to run the Windows legs within a Docker image which may be Nano Server based.  Because of this, I am going to leave the comment as is.

mthalman (approved):

mthalman (approved):

PR #641, opened 2020-09-15
Delete everything but vswhere from VS installers directory

### LINKED ISSUE #632: VS installer deletes vswhere from installation
From the 4.8 SDK Dockerfile:

https://github.com/microsoft/dotnet-framework-docker/blob/5d8e932818fcf188618f18eab4c978cbd1e6a0e6/src/sdk/4.8/windowsservercore-ltsc2019/Dockerfile#L34

Presumably this is an image-size optimization, but it deletes vswhere.exe, which causes issues in tools that rely upon the documented behavior for vswhere: 
> Starting with Visual Studio 15.2 (26418.1 Preview) vswhere.exe is installed in %ProgramFiles(x86)%\Microsoft Visual Studio\Installer (use %ProgramFiles% in a 32-bit program prior to Windows 10). This is a fixed location that will be maintained.

_(From https://github.com/Microsoft/vswhere/wiki/Installing)_

I recognize that this image builds VS's build tools into a static location, but I'll have to bring in an outside vswhere.exe for the third-party tools that expect to find this tool, because I don't see it anywhere else in the image.

mthalman:
Fixes #632

### DIFF
--- a/src/sdk/3.5/windowsservercore-1903/Dockerfile
+++ b/src/sdk/3.5/windowsservercore-1903/Dockerfile
@@ -31,7 +31,8 @@ RUN `
     && del vs_BuildTools.exe `
     `
     #Cleanup
-    && rmdir /S /Q "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer" `
+    && (for /D %i in ("%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\*") do rmdir /S /Q "%i") `
+    && (for %i in ("%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\*") do if not "%~nxi" == "vswhere.exe" del "%~i") `
     && powershell Remove-Item -Force -Recurse "%TEMP%\*" `
     && rmdir /S /Q "%ProgramData%\Package Cache"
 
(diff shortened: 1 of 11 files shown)


MichaelSimons (commented):
Why only change the 4.8 images?  Should the 3.5 images also be updated?

Thoughts on creating a unit test to ensure vswhere is runnable?

mthalman:
> Why only change the 4.8 images? Should the 3.5 images also be updated?

Just an oversight.  Need some Dockerfile templates.  :)

> Thoughts on creating a unit test to ensure vswhere is runnable?

I don't think it's that high of a priority.  Considering it's a single file executable, there's not really much we can do to break it.  And there's plenty of other executables that get installed that we don't test.

MichaelSimons (approved):
> > Thoughts on creating a unit test to ensure vswhere is runnable?
> 
> I don't think it's that high of a priority.  Considering it's a single file executable, there's not really much we can do to break it.  And there's plenty of other executables that get installed that we don't test.

That's fair but keep in mind another motivation for some form of test is to simply ensure the presence of the exe and that it doesn't regress so to speak.

mthalman:
> > > Thoughts on creating a unit test to ensure vswhere is runnable?
> > 
> > 
> > I don't think it's that high of a priority.  Considering it's a single file executable, there's not really much we can do to break it.  And there's plenty of other executables that get installed that we don't test.
> 
> That's fair but keep in mind another motivation for some form of test is to simply ensure the presence of the exe and that it doesn't regress so to speak.

I forgot that we already have an existing test which downloads vswhere and verifies it runs.  I've simply updated it to reference vswhere.exe on disk instead of downloading it.

PR #649, opened 2020-10-05
Skip pre-build tests on Windows containers

mthalman:
Fixes dotnet/dotnet-docker#2261

### DIFF
--- a/build-and-test.ps1
+++ b/build-and-test.ps1
@@ -27,7 +27,6 @@ else {
 
 if ($Repos.Count -eq 0) {
     $Path = $null
-    $testCategories = @()
 }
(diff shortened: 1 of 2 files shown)


MichaelSimons (approved):

mthalman:
I made another update because this wasn't a complete solution.  The problem is that the build script will override the default `TestCategories` parameter by setting it to an empty array.  This defaults to running all the tests, including the pre-build tests.  I've updated the build script to optionally include the `TestCategories` parameter only if it needs to be set.  This allows the default values from the test script to take effect.

MichaelSimons (approved):

PR #677, opened 2020-11-18
Change branding of Windows Server, version 2009 to 20H2

### LINKED ISSUE #671: Change Windows 2009 tagging to 20H2
While "2009" followed the same convention as previous releases, the official name for it is 20H2.

mthalman:
The only things I didn't change was the names of the demands for the build agents and the name of the 3.5 zip file in blob storage: `microsoft-windows-netfx3-2009.zip`.  It would be a breaking change to rename the zip file.  The alternative would be to make a copy of it.  It doesn't look like there's a way to do symbolic links with Azure blob storage.

Fixes #671


manifest.samples.json
```diff
@@ -57,9 +57,12 @@
             {
               "dockerfile": "samples/dotnetapp",
               "os": "windows",
-              "osVersion": "windowsservercore-2009",
+              "osVersion": "windowsservercore-20H2",
               "tags": {
-                "dotnetapp-windowsservercore-2009": {}
+                "dotnetapp-windowsservercore-20H2": {},
+                "dotnetapp-windowsservercore-2009": {
```
MichaelSimons, line 63: The same comment I made in the .NET Core change applies here.  I question the need to maintain the 2009 tags in the samples repo.

mthalman: Fixed

MichaelSimons (approved):

jeanfrancoislarente:
Hello @mthalman ,

We're in limbo at the moment with our automation and the discrepancy between servercore and framework tags (we would have to introduce a workaround with the current tagging).

Do you have an ETA of when this would be merged and made publicly available?

Thanks very much!

mthalman:
> We're in limbo at the moment with our automation and the discrepancy between servercore and framework tags (we would have to introduce a workaround with the current tagging).

@jeanfrancoislarente -
Are you referring to the differences of what tags are available for the .NET Framework images versus what's available for the base Windows Server Core images?  If so, Windows does still provide the `2009` tags (i.e. this tag still exists: `mcr.microsoft.com/windows/servercore:2009-amd64`).  So you can still use consistent tags.  These changes will be rolled out as part of the next release on Patch Tuesday (Dec. 8).

jeanfrancoislarente:
Thanks for the response @mthalman - we'll wait for December 8th patch!

PR #701, opened 2021-01-07
Include NuGet 5.8 in SDK as additional version

### LINKED ISSUE #678: Nuget 5.7 incompatible with VS 2019.8
Looks like currently mcr.microsoft.com/dotnet/framework/sdk:4.8 includes msbuild from Visual Studio 2019 Update 8 and NuGet 5.7. They are not compatible with each other - attempt to build project with VS msbuild after restoring with NuGet will result in errors like:
error NETSDK1005: Assets file '...\obj\project.assets.json' doesn't have a target for '...'. Ensure that restore has run and that you have included '...' in the TargetFrameworks for your project.

mthalman:
Fixes #678

These changes include NuGet 5.8 in the SDK in order to address the incompatibilities introduced by VS 16.8 build tools with older versions of NuGet.  NuGet 5.8 is included as an additional installation that developers can optionally use in order to workaround this incompatibility.  The existing versions of NuGet in the SDK is not being changed at this time in order to avoid potential breaking changes associated with such an upgrade.

### DIFF
--- a/eng/models/NuGetInfo.cs
+++ b/eng/models/NuGetInfo.cs
@@ -14,5 +14,8 @@ namespace Microsoft.DotNet.Framework.Models
 
         [JsonProperty(Required = Required.Always)]
         public string NuGetClientVersion { get; set; } = String.Empty;
+
+        [JsonProperty(Required = Required.Always)]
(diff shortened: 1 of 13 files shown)


MichaelSimons (approved):
LGTM pending feedback from the NuGet team.

Issue #616, opened 2020-07-21
Templates for Dockerfiles


mthalman:
Now that https://github.com/dotnet/dotnet-docker/issues/1933 has the necessary infrastructure in place, defining templates for the .NET Fx Dockerfiles can be considered.  There should be an investigation to determine where there is commonality between the Dockerfiles and whether its reasonable to define templates.  Benefits of having templates, or at least going through the process, include being able to identify potential inconsistencies that were unintentional.

PR #705, opened 2021-01-08
Add infra to generate wcf Dockerfiles from templates

MichaelSimons:
This is the first phase of introducing Dockerfile templates.  The other variants such as aspnet, runtime, and sdk will be added if deemed valuable in subsequent PRs.

Related to #616

### DIFF
--- a/eng/common/templates/variables/docker-images.yml
+++ b/eng/common/templates/variables/docker-images.yml
@@ -1,5 +1,5 @@
 variables:
-  imageNames.imageBuilder: mcr.microsoft.com/dotnet-buildtools/image-builder:20210108104530
+  imageNames.imageBuilder: mcr.microsoft.com/dotnet-buildtools/image-builder:943088
   imageNames.imageBuilder.withrepo: imagebuilder-withrepo:$(Build.BuildId)-$(System.JobId)
   imageNames.testRunner: mcr.microsoft.com/dotnet-buildtools/prereqs:debian-stretch-slim-docker-testrunner-690072
   imageNames.testRunner.withrepo: testrunner-withrepo:$(Build.BuildId)-$(System.JobId)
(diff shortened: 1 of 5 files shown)


eng/common/templates/variables/docker-images.yml
```diff
@@ -1,5 +1,5 @@
 variables:
-  imageNames.imageBuilder: mcr.microsoft.com/dotnet-buildtools/image-builder:939194
+  imageNames.imageBuilder: mcr.microsoft.com/dotnet-buildtools/image-builder:20210108104530
```
MichaelSimons, line 2: This needs to be updated with an official build with https://github.com/dotnet/docker-tools/pull/727.

eng/dockerfile-templates/wcf/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+{{if OS_VERSION_NUMBER = "ltsc2016" || OS_VERSION_NUMBER = "ltsc2019" || OS_VERSION_NUMBER = "1903" || OS_VERSION_NUMBER = "1909" || OS_VERSION_NUMBER = "2004"
+:RUN Add-WindowsFeature NET-WCF-TCP-Activation45; `
+    Add-WindowsFeature NET-WCF-HTTP-Activation45; `
+    Add-WindowsFeature Web-WebSockets
+^else
+:RUN dism /Online /Quiet /Enable-Feature /All /FeatureName:WCF-HTTP-Activation45 /FeatureName:WCF-TCP-Activation45 /FeatureName:IIS-WebSockets
+}}
+# Enable net.tcp protocol for default web site on IIS
+{{if OS_VERSION_NUMBER = "ltsc2016" || OS_VERSION_NUMBER = "ltsc2019" || OS_VERSION_NUMBER = "1903" || OS_VERSION_NUMBER = "1909" || OS_VERSION_NUMBER = "2004"
+:RUN windows\system32\inetsrv\appcmd.exe set app 'Default Web Site/' /enabledProtocols:"http,net.tcp"
```
mthalman, line 16: This seems like a non-functional difference compared to 20H2 that could simply be updated for these versions to provide consistency with 20H2.

MichaelSimons: The trouble is that pre 20H2, the SHELL is set to PowerShell.  I am not finding a syntax that works with both shells that makes sense going forward.  I am open to suggestions incase I missed something obvious.

mthalman: Ok, I see now.

mthalman (approved):

PR #711, opened 2021-01-11
Upgrade System.CommandLine version

### LINKED ISSUE #707: Replace dependency on System.CommandLine
The update-dependencies tool has a dependency on an older, unsupported version of the `System.CommandLine` NuGet package which had existed in a `dotnet.myget.org` feed but that feed has now been deprecated and no longer available.  For that reason, restoring packages for the project will fail with an error similar to the following:

```
/update-dependencies/update-dependencies.csproj : warning NU1603: update-dependencies depends on System.CommandLine (>= 0.1.0-e170603-2) but System.CommandLine 0.1.0-e170603-2 was not found. An approximate best match of System.CommandLine 0.3.0-alpha.20054.1 was resolved.
/update-dependencies/Options.cs(27,13): error CS0246: The type or namespace name 'ArgumentSyntax' could not be found (are you missing a using directive or an assembly reference?) [/update-dependencies/update-dependencies.csproj]
/update-dependencies/Options.cs(27,40): error CS0103: The name 'ArgumentSyntax' does not exist in the current context [/update-dependencies/update-dependencies.csproj]
```

The project needs to take a new dependency on a supported version of `System.CommandLine`.

mthalman:
Removes the use of an older, unsupported version of `System.CommandLine` that is no longer available from the deprecated `dotnet.myget.org` feed.  Replaces it with the new version of `System.CommandLine` from https://github.com/dotnet/command-line-api. This has a new API model that the code has been refactored to consume.

Fixes #707


eng/update-dependencies/Options.cs
```diff
@@ -31,96 +35,44 @@ public class Options
 
         public string? GitHubUser { get; private set; }
 
-        public string NuGetInfoPath { get; private set; } = "eng/nuget-info.json";
+        public string NuGetInfoPath { get; private set; } = DefaultNuGetInfoPath;
```
MichaelSimons, line 38: Why is it necessary to specify a default for the property here as well as the GetCliSymbols method?

mthalman: I guess it's not needed.

MichaelSimons (approved):

PR #712, opened 2021-01-12
Add sdk Dockerfile templates

MichaelSimons:
Related to #616


mthalman (approved):
I've been waiting for this day.  Looks great.  I like how this makes use of manifest.versions.json to replace the other json files.

PR #743, opened 2021-03-09
Servicing updates for 3B release

mthalman:
Related to https://github.com/microsoft/dotnet-framework-docker/issues/741


MichaelSimons (approved):

mthalman:
Reverted the 2019 changes because that's meant for .NET servicing.  There is no .NET servicing.  So the only changes that should be applied are for 2016.

Issue #782, opened 2021-06-14
Remove redundant patching in Dockerfiles


mthalman:
Windows is now servicing the .NET Framework version that is installed in their base images starting with version Server 2019 and newer.  Because of that the .NET Framework Dockerfiles do not need to explicitly install the patch; doing so is just redundant because it's already been installed.  This removal of the patching from the Dockerfile had already been done in the last release but only for Server 2019.  The Dockerfiles for 2004 and 20H2 should also be updated in the same manner.

Just to be clear on this, I'll provide an example.  In the Windows 20H2 image, you can see that .NET Fx 4.8 has been patched:

```console
❯ docker run --rm mcr.microsoft.com/windows/servercore:20H2 powershell get-hotfix

Source        Description      HotFixID      InstalledBy          InstalledOn
------        -----------      --------      -----------          -----------
C47AE93B5D9F  Update           KB4601050                          6/6/2021 1...
C47AE93B5D9F  Update           KB4562830                          11/18/2020...
C47AE93B5D9F  Security Update  KB4570334                          11/18/2020...
C47AE93B5D9F  Security Update  KB5003637                          6/6/2021 1...
C47AE93B5D9F  Update           KB5003503                          6/6/2021 1...
```

The first update listed, [KB4601050](https://support.microsoft.com/topic/february-9-2021-kb4601050-cumulative-update-for-net-framework-3-5-and-4-8-for-windows-10-version-2004-windows-server-version-2004-windows-10-version-20h2-and-windows-server-version-20h2-7d61f856-72ce-7f4c-7e43-76f8aadf2aca), is the latest 4.8 patch.

And the Dockerfile for 4.8 on 20H2 is unnecessarily installing that patch:

https://github.com/microsoft/dotnet-framework-docker/blob/d4095f39459c58f1951abbf21cf0a2b95458fac4/src/runtime/4.8/windowsservercore-20H2/Dockerfile#L11-L17

PR #785, opened 2021-06-17
Remove redundant patch installation for 4.8 on 2004 and 20H2

mthalman:
Fixes https://github.com/microsoft/dotnet-framework-docker/issues/782

### DIFF
--- a/src/runtime/4.8/windowsservercore-2004/Dockerfile
+++ b/src/runtime/4.8/windowsservercore-2004/Dockerfile
@@ -9,8 +9,6 @@ ENV `
 
 RUN `
     # Ngen top of assembly graph to optimize a set of frequently used assemblies
-    && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" `
+    %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"
     # To optimize 32-bit assemblies, uncomment the next line
     # && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" `
-    && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen update `
-    && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen update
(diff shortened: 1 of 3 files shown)


src/runtime/4.8/windowsservercore-2004/Dockerfile
```diff
@@ -8,14 +8,6 @@ ENV `
...
 
 RUN `
-    # Apply latest patch
-    curl -fSLo patch.msu http://download.windowsupdate.com/d/msdownload/update/software/secu/2021/01/windows10.0-kb4601050-x64-ndp48_04c8736f250c06b3787d7d9ec366f011c6f8ca91.msu `
-    && mkdir patch `
-    && expand patch.msu patch -F:* `
-    && del /F /Q patch.msu `
-    && dism /Online /Quiet /Add-Package /PackagePath:C:\patch\windows10.0-kb4601050-x64-ndp48.cab `
-    && rmdir /S /Q patch `
-    `
```
MichaelSimons, line 18: Is the `&& %windir%\Microsoft.NET\Framework\v4.0.30319\ngen update` call necessary anymore?  I am wondering if this was only necessary because of the patch application?

mthalman: Fixed, PTAL

src/runtime/4.8/windowsservercore-2004/Dockerfile
```diff
@@ -8,17 +8,7 @@ ENV `
...
     # Ngen top of assembly graph to optimize a set of frequently used assemblies
-    && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" `
+    %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35"
     # To optimize 32-bit assemblies, uncomment the next line
     # && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen install "Microsoft.PowerShell.Utility.Activities, Version=3.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" `
-    && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen update `
```
MichaelSimons, line 23: This looks good.  Given the number of issues in the past with the Windows base images, do you think it would be worthwhile to have a unit test that validates the ngen queue is empty?  If so I don't think it is necessary for this PR and could be tracked separately.

MichaelSimons (approved):

src/runtime/4.8/windowsservercore-2004/Dockerfile
mthalman: We already do:

https://github.com/microsoft/dotnet-framework-docker/blob/d4095f39459c58f1951abbf21cf0a2b95458fac4/tests/Microsoft.DotNet.Framework.Docker.Tests/RuntimeOnlyImageTests.cs#L52

PR #786, opened 2021-06-18
Update the Roslyn to 3.6.0

### LINKED ISSUE #774: Consider updating Roslyn binaries
I noticed that the Roslyn package we use is super old.

https://github.com/microsoft/dotnet-framework-docker/blob/66cfc9536db13fa27b82c2bc33fd916e5f583366/eng/dockerfile-templates/aspnet/Dockerfile#L18

We're using a version from 2018. https://www.nuget.org/packages/Microsoft.Net.Compilers/

This package should probably get updated every time we release (as a checklist item). I'd recommend looking at all the content we curl down. It looks like there are some other assets that might be stale.

/cc @jaredpar

Benjie-Liu:
Fixes #774

### DIFF
--- a/eng/dockerfile-templates/aspnet/Dockerfile
+++ b/eng/dockerfile-templates/aspnet/Dockerfile
@@ -15,10 +15,10 @@ RUN dism /Online /Quiet /Enable-Feature /All /FeatureName:IIS-WebServerRole {{if
 {{if PRODUCT_VERSION != "3.5"
 :
 # Install Roslyn compilers
-RUN curl -fSLo microsoft.net.compilers.2.9.0.zip https://api.nuget.org/packages/microsoft.net.compilers.2.9.0.nupkg `
+RUN curl -fSLo microsoft.net.compilers.3.6.0.zip https://api.nuget.org/packages/microsoft.net.compilers.3.6.0.nupkg `
     && mkdir C:\RoslynCompilers `
-    && tar -C C:\RoslynCompilers -zxf microsoft.net.compilers.2.9.0.zip `
-    && del microsoft.net.compilers.2.9.0.zip `
+    && tar -C C:\RoslynCompilers -zxf microsoft.net.compilers.3.6.0.zip `
+    && del microsoft.net.compilers.3.6.0.zip `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\csc.exe /ExeConfig:C:\RoslynCompilers\tools\csc.exe `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\vbc.exe /ExeConfig:C:\RoslynCompilers\tools\vbc.exe `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\VBCSCompiler.exe /ExeConfig:C:\RoslynCompilers\tools\VBCSCompiler.exe
--- a/eng/dockerfile-templates/aspnet/Dockerfile
+++ b/eng/dockerfile-templates/aspnet/Dockerfile
@@ -14,7 +14,7 @@ RUN dism /Online /Quiet /Enable-Feature /All /FeatureName:IIS-WebServerRole {{if
     && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen update
 {{if PRODUCT_VERSION != "3.5"
 :
-# Install Roslyn compilers
+# Install 2.9.0 Roslyn compilers
 RUN curl -fSLo microsoft.net.compilers.2.9.0.zip https://api.nuget.org/packages/microsoft.net.compilers.2.9.0.nupkg `
     && mkdir C:\RoslynCompilers `
     && tar -C C:\RoslynCompilers -zxf microsoft.net.compilers.2.9.0.zip `
--- a/eng/dockerfile-templates/aspnet/Dockerfile
+++ b/eng/dockerfile-templates/aspnet/Dockerfile
@@ -15,15 +15,24 @@ RUN dism /Online /Quiet /Enable-Feature /All /FeatureName:IIS-WebServerRole {{if
 {{if PRODUCT_VERSION != "3.5"
 :
 # Install Roslyn compilers
-RUN curl -fSLo microsoft.net.compilers.3.6.0.zip https://api.nuget.org/packages/microsoft.net.compilers.3.6.0.nupkg `
+RUN curl -fSLo microsoft.net.compilers.2.9.0.zip https://api.nuget.org/packages/microsoft.net.compilers.2.9.0.nupkg `
     && mkdir C:\RoslynCompilers `
-    && tar -C C:\RoslynCompilers -zxf microsoft.net.compilers.3.6.0.zip `
-    && del microsoft.net.compilers.3.6.0.zip `
+    && tar -C C:\RoslynCompilers -zxf microsoft.net.compilers.2.9.0.zip `
+    && del microsoft.net.compilers.2.9.0.zip `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\csc.exe /ExeConfig:C:\RoslynCompilers\tools\csc.exe `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\vbc.exe /ExeConfig:C:\RoslynCompilers\tools\vbc.exe `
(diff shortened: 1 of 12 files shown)


mthalman (commented):
In addition to updating the template, the Dockerfiles need to be regenerated from the template.  Please read the instructions for [updating Dockerfiles](https://github.com/microsoft/dotnet-framework-docker/blob/main/CONTRIBUTING.md#editing-dockerfiles).

mthalman:
@Benjie-Liu - Please sign the [CLA](https://github.com/microsoft/dotnet-framework-docker/pull/786#issuecomment-865696184).

MichaelSimons:
Can updating the https://github.com/microsoft/dotnet-framework-docker/blob/main/.github/ISSUE_TEMPLATE/releases/patch-tuesday-release.md to include a Roslyn update step be included in this PR?

HongGit (approved):

StephenMolloy:
Do we understand the use case of this Roslyn package in our docker containers? This package is from the compiler team... it is not the RoslynCodeDomProvider from the ASP.Net team which would be included as part of the web app being deployed in the container. That package is logically coupled with a particular version of Microsoft.Net.Compilers, which is why its versioning changed in the latest release to reflect the MS.Net.Compilers version it is tied to. In these Dockerfiles, we do set the environment variable that RoslynCodeDomProvider looks for to locate the roslyn tools. So we are effecitvely changing the Roslyn version out from under any app that uses the RoslynCodeDomProvider package, whether they are still on the 2.0.1 train or they have updated to the 3.6 package.

Now, I'm not aware of any compat issues that might arise from using the 2.0.1 RoslynCodeDomProvider package with the 3.6 compilers, or vice versa (3.6 codedom provider with the 2.9 compilers), but we haven't really tested that scenario, and I believe mixing compiler versions with our codedom provider has always been an "at your own risk" sort of thing. With the way we are doing things here, we are most likely creating that situation for folks who aren't paying attention.

mthalman:
@HongGit - Can you comment on the [question](https://github.com/microsoft/dotnet-framework-docker/pull/786#issuecomment-866575536) from @StephenMolloy?  It'd be good to have this understood and resolved before next month's release.

Benjie-Liu:
Update the Dockerfiles as Steve and Hong discussed: Keep up the 2.9 compilers as before, add the Roslyn 3.6.0 installation and update the ROSLYN_COMPILER_LOCATION to the latest version C:\RoslynCompilers-3.6.0\tools.

eng/dockerfile-templates/aspnet/Dockerfile.pre20H2
```diff
@@ -17,17 +17,30 @@ RUN Add-WindowsFeature Web-Server; `
...
+    Remove-Item C:\microsoft.net.compilers.3.6.0.zip -Force; `
+    &$Env:windir\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\csc.exe /ExeConfig:C:\RoslynCompilers-3.6.0\tools\csc.exe | `
+    &$Env:windir\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\vbc.exe /ExeConfig:C:\RoslynCompilers-3.6.0\tools\vbc.exe | `
+    &$Env:windir\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\VBCSCompiler.exe /ExeConfig:C:\RoslynCompilers-3.6.0\tools\VBCSCompiler.exe{{if OS_VERSION_NUMBER != "2004": | `
+    &$Env:windir\Microsoft.NET\Framework\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\csc.exe /ExeConfig:C:\RoslynCompilers-3.6.0\tools\csc.exe | `
+    &$Env:windir\Microsoft.NET\Framework\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\vbc.exe /ExeConfig:C:\RoslynCompilers-3.6.0\tools\vbc.exe | `
+    &$Env:windir\Microsoft.NET\Framework\v4.0.30319\ngen install C:\RoslynCompilers-3.6.0\tools\VBCSCompiler.exe  /ExeConfig:C:\RoslynCompilers-3.6.0\tools\VBCSCompiler.exe}}
+
+
```
mthalman, line 42: Remove extra line

eng/dockerfile-templates/aspnet/Dockerfile
```diff
@@ -15,15 +15,24 @@ RUN dism /Online /Quiet /Enable-Feature /All /FeatureName:IIS-WebServerRole {{if
 {{if PRODUCT_VERSION != "3.5"
 :
 # Install Roslyn compilers
```
mthalman: `# Install 2.9.0 Roslyn compilers`

eng/dockerfile-templates/aspnet/Dockerfile.pre20H2
```diff
@@ -17,17 +17,30 @@ RUN Add-WindowsFeature Web-Server; `
 
 # Install Roslyn compilers
```
mthalman: `# Install 2.9.0 Roslyn compilers`

StephenMolloy (approved):

eng/dockerfile-templates/aspnet/Dockerfile.pre20H2
Benjie-Liu: Removed the extra line.

eng/dockerfile-templates/aspnet/Dockerfile
Benjie-Liu: Updated the comments: # Install 2.9.0 Roslyn compilers

eng/dockerfile-templates/aspnet/Dockerfile.pre20H2
Benjie-Liu: Updated.

MichaelSimons:
As part of this work I would like to see a policy written down for the Roslyn versioning storing.  What version(s) will be included and how/when it will get serviced/updated going forward.

mthalman (commented):
In addition to the [aspnet/Dockerfile template file](https://github.com/microsoft/dotnet-framework-docker/blob/main/eng/dockerfile-templates/aspnet/Dockerfile), the [aspnet/Dockerfile.pre20H2 template file](https://github.com/microsoft/dotnet-framework-docker/blob/main/eng/dockerfile-templates/aspnet/Dockerfile.pre20H2) also needs to be updated with the same changes.

eng/dockerfile-templates/aspnet/Dockerfile
```diff
@@ -23,7 +23,16 @@ RUN curl -fSLo microsoft.net.compilers.2.9.0.zip https://api.nuget.org/packages/
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\vbc.exe /ExeConfig:C:\RoslynCompilers\tools\vbc.exe `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen install C:\RoslynCompilers\tools\VBCSCompiler.exe /ExeConfig:C:\RoslynCompilers\tools\VBCSCompiler.exe
 
-ENV ROSLYN_COMPILER_LOCATION=C:\RoslynCompilers\tools
+# Install 3.6.0 Roslyn compilers
```
mthalman, line 26: Remove trailing space at end of line and regenerate affected Dockerfiles.

mthalman (approved):

eng/dockerfile-templates/aspnet/Dockerfile
```diff
@@ -23,7 +23,16 @@ RUN curl -fSLo microsoft.net.compilers.2.9.0.zip https://api.nuget.org/packages/
...
+RUN curl -fSLo microsoft.net.compilers.3.6.0.zip https://api.nuget.org/packages/microsoft.net.compilers.3.6.0.nupkg `
```
jaredpar, line 27: This package is deprecated, please use microsoft.net.compilers.toolset instead 

From the NuPkg:

> Note: This package is deprecated. Please use Microsoft.Net.Compilers.Toolset instead

jaredpar: Fine for this PR, but in the future we will stop producing new versions of this package.

jaredpar:
> So we are effecitvely changing the Roslyn version out from under any app that uses the RoslynCodeDomProvider package, whether they are still on the 2.0.1 train or they have updated to the 3.6 package.

In general that should be fine. The compilers have an incredibly high compatibility bar and our recommendation to customers is always "please upgrade to latest". If you hit errors that would be the rare case, not the norm. Yes changes do exist from version to version but they are generally in decently obscure corner cases

PR #801, opened 2021-08-02
Prevent generation of ASP.NET Core developer certificate

mthalman:
Related to https://github.com/microsoft/dotnet-framework-docker/issues/799


eng/dockerfile-templates/sdk/Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN mkdir "%ProgramFiles%\NuGet\latest" `
     && curl -fSLo "%ProgramFiles%\NuGet\nuget.exe" https://dist.nuget.org/win-x86-commandline/v%NUGET_VERSION%/nuget.exe `
     && mklink "%ProgramFiles%\NuGet\latest\nuget.exe" "%ProgramFiles%\NuGet\nuget.exe"
 
+# Do not generate certificate
```
MichaelSimons, line 32: At some point, I think we should consider a pattern where the ENVs are defined in a single place as done in the .NET images.  The primary goal is to improve the readability of discovering what ENVs are defined.

MichaelSimons (commented):
I was expecting the [SDK's ENV tests](https://github.com/microsoft/dotnet-framework-docker/blob/main/tests/Microsoft.DotNet.Framework.Docker.Tests/SdkOnlyImageTests.cs#L85) to be update to validate the new ENV.

MichaelSimons (approved):

PR #815, opened 2021-09-02
Consolidate ENV instructions in sdk Dockerfiles

### LINKED ISSUE #802: Consider consolidating ENV instructions
The SDK Dockerfiles ([example](https://github.com/microsoft/dotnet-framework-docker/blob/main/src/sdk/4.8/windowsservercore-2004/Dockerfile)) currently have two separate `ENV` instructions and a third one will be added with https://github.com/microsoft/dotnet-framework-docker/pull/801.  These create separate layers within the resulting image.  There's no requirement for them to be in separate layers.  It would be cleaner to consolidate all the environment variable declarations in one instruction.

mthalman:
Fixes #802


MichaelSimons (commented):
Nice!

PR #888, opened 2022-02-08
Fix build issues

mthalman:
Needing to revert back the version of Image Builder because some of the changes to support arch "variant" aren't working on the 2016 machines.

Also fixing the Dockerfile generation for 2022.

### DIFF
--- a/src/runtime/4.8/windowsservercore-ltsc2022/Dockerfile
+++ b/src/runtime/4.8/windowsservercore-ltsc2022/Dockerfile
@@ -10,7 +10,7 @@ ENV `
 
 RUN `
     # Apply latest patch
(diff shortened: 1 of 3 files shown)


src/runtime/4.8/windowsservercore-ltsc2022/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV `
 
 RUN `
     # Apply latest patch
-    curl -fSLo patch.msu http://download.windowsupdate.com/d/msdownload/update/software/updt/2021/12/windows10.0-kb5009470-x64-ndp48_bb024317f391de318cc42f12641da9b3253011cf.msu `
+    && curl -fSLo patch.msu http://download.windowsupdate.com/d/msdownload/update/software/updt/2021/12/windows10.0-kb5009470-x64-ndp48_bb024317f391de318cc42f12641da9b3253011cf.msu `
```
MichaelSimons, line 13: This change doesn't look right.

MichaelSimons (approved):

PR #952, opened 2022-05-11
Ngen workaround for arm64 version of MSBuild and TestHost

mthalman:
Related to https://github.com/dotnet/release/issues/410

Execution of `ngen update` in the sdk Dockerfiles is failing because an arm64 version of MSBuild is mistakenly queued for ngen. This results in the errors:

`Failed to compile C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin\arm64\MSBuild.exe because this image is not a valid Win32 application.`

`An attempt was made to load a program with an incorrect format. (Exception from HRESULT: 0x8007000B)`

As a workaround, I've uninstalled it from the ngen queue.

Update: This also applies to several variations of testhost.exe from VS TestAgent.


mthalman:
Here's the tracking issue I've created for MSBuild.exe: https://github.com/microsoft/dotnet-framework-docker/issues/953

mthalman:
Converted this to a draft while I investigate a 32-bit ngen issue.

mthalman:
Ok, so looks like this testhost.exe is also a problem here for 32-bit ngen. Updated the Dockerfiles for that as well.

mthalman:
Here is the tracking issue for testhost.exe: https://github.com/microsoft/dotnet-framework-docker/issues/954

MichaelSimons (approved):

PR #975, opened 2022-07-20
VS 17.3 updates

mthalman:
Updates the VS build tools to 17.3.

For now, this is targeting the Preview version of 17.3 since it's not yet released. This will be updated before merging on release day.

To allow for things to work against a Preview version, I needed to make the following changes:
* Explicitly set the `--install-path` option for the installer. Otherwise, the installer would install to the `%ProgramFiles(x86)%\Microsoft Visual Studio\2022\Preview` path for both Test Agent and Build Tools, causing a conflict.
* Logic to parse preview version of .NET installation in the tests.
* Logic to gracefully handle preview versions when using vswhere utility.

The upgrade to 17.3 also allowed the removal of some GAC workarounds now that the underlying issues have been fixed.

### DIFF
--- a/eng/dockerfile-templates/sdk/Dockerfile
+++ b/eng/dockerfile-templates/sdk/Dockerfile
@@ -42,13 +42,14 @@ RUN mkdir "%ProgramFiles%\NuGet\latest" `
 RUN `
     # Install VS Test Agent
     curl -fSLo vs_TestAgent.exe {{VARIABLES["vs|testAgentUrl"]}} `
-    && start /w vs_TestAgent --quiet --norestart --nocache --wait `
+    && start /w vs_TestAgent --quiet --norestart --nocache --wait --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\TestAgent" `
     && powershell -Command "if ($err = dir $Env:TEMP -Filter dd_setup_*_errors.log | where Length -gt 0 | Get-Content) { throw $err }" `
     && del vs_TestAgent.exe `
     `
     # Install VS Build Tools
     && curl -fSLo vs_BuildTools.exe {{VARIABLES["vs|buildToolsUrl"]}} `
     && start /w vs_BuildTools @^ `
+        --installPath "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\BuildTools" @^ `
         --add Microsoft.Component.ClickOnce.MSBuild @^ `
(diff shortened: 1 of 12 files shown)


eng/dockerfile-templates/sdk/Dockerfile
```diff
@@ -65,22 +66,10 @@ RUN `
...
-    && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\TestAgent\Common7\IDE\Extensions\TestPlatform\testhost.net47.arm64.exe" `
-    && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\TestAgent\Common7\IDE\Extensions\TestPlatform\testhost.net471.arm64.exe" `
-    && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\TestAgent\Common7\IDE\Extensions\TestPlatform\testhost.net472.arm64.exe" `
-    && %windir%\Microsoft.NET\Framework\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft Visual Studio\2022\TestAgent\Common7\IDE\Extensions\TestPlatform\testhost.net48.arm64.exe" `
-    `}}
+    `
     # Workaround for issues with 64-bit ngen
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\SecAnnotate.exe" `
     && %windir%\Microsoft.NET\Framework64\v4.0.30319\ngen uninstall "%ProgramFiles(x86)%\Microsoft SDKs\Windows\v10.0A\bin\NETFX 4.8 Tools\WinMDExp.exe" `
```
MichaelSimons, line 72: Do you know how come they fixes some but not all of the bad ngen queued items?

MichaelSimons (commented):
Looks good

eng/dockerfile-templates/sdk/Dockerfile
mthalman: All of the newly introduced issues were fixed. There are separate teams that are responsible for each of the categories of files that were mistakenly included.

PR #983, opened 2022-08-09
Delete Windows Server 20H2 Dockerfiles

mthalman:
Deleting the Dockerfiles based on Windows Server 20H2 because it is EOL as of today.


MichaelSimons (approved):
LGTM - always nice to remove code!

PR #1018, opened 2022-10-25
Readme updates

mthalman:
Updates the header names to align with the [names specified for the MAR portal](https://eng.ms/docs/cloud-ai-platform/azure-core/azure-management-and-platforms/containers-bburns/microsoft-container-registry/microsoft-artifact-registry-also-known-as-microsoft-container-registry-docs/onboardingdocs/documentationstandards).

Also adds WS 2022 for .NET Fx 3.5 to the version compatibility table.


MichaelSimons (approved):
Is there going to be a follow-up PR to re-order the sections for the MCR readmes?

PR #1026, opened 2022-11-17
Use Dockerfile paths to determine which test data is in scope

mthalman:
This updates the test project to reflect the changes that were made in https://github.com/dotnet/dotnet-docker/pull/4207 and to conform to the new test script contract defined by https://github.com/dotnet/docker-tools/pull/1082. This means the test project no longer filters tests by product version but rather by Dockerfile path.

This change made it very cumbersome to attempt to adapt the existing `SkippableTheoryAttribute` class to use the Dockerfile path data in order to determine which tests to skip. So instead of doing that, I got rid of it entirely and went with a different approach. Each of the test projects defined its own set of test data. This data was largely duplicated amongst the classes except for WCF and SDK tests which needed to limit which versions they tested (hence the need for `SkippableTheoryAttribute`).

I consolidated all of that test data into one set which all the test classes use. Then I made use of the `Xunit.SkippableFact` NuGet package. This provides its own `SkippableTheoryAttribute` that, when used, allows the code to throw a `SkipException` to indicate that the test should be skipped. It also provides some helper methods that will throw a `SkipException` based on a condition. This then allowed me to reimplement the skip logic based on version in a much more straightforward manner.


tests/Microsoft.DotNet.Framework.Docker.Tests/SdkOnlyImageTests.cs
```diff
@@ -15,36 +15,31 @@ namespace Microsoft.DotNet.Framework.Docker.Tests
...
         protected override string ImageType => "sdk";
 
-        public static IEnumerable<object[]> GetImageData()
-        {
-            return ImageTestHelper.ApplyImageDataFilters(s_imageData);
-        }
+        public static IEnumerable<object[]> GetImageData() =>
+            ImageTestHelper.ApplyImageDataFilters(TestData.GetImageData(), "sdk");
+
+        private static bool IsSkippable(ImageDescriptor imageDescriptor) =>
```
MichaelSimons, line 28: Would it make sense to express this with via logic in the ApplyImageDataFilters method?  Anytime the SDK version doesn't align with the runtime version a filter should be applied.  This would help cleanup the tests here  in the sense they wouldn't have to be declared skippable.

tests/Microsoft.DotNet.Framework.Docker.Tests/WcfImageTests.cs
```diff
@@ -8,41 +8,36 @@
...
         }
 
-        protected override string ImageType => "wcf";
+        protected override string ImageType => ImageTypes.Wcf;
+
+        public static IEnumerable<object[]> GetImageData() =>
+            ImageTestHelper.ApplyImageDataFilters(TestData.GetImageData(), ImageTypes.Wcf);
 
-        [SkippableTheory("3.5")]
+        private static bool IsSkippable(ImageDescriptor imageDescriptor) => imageDescriptor.Version == "3.5";
```
MichaelSimons, line 24: Why not just apply this filter to the GetImageData and avoid all of the calls to `Skip.If(...)`

```
public static IEnumerable<object[]> GetImageData() =>
            ImageTestHelper.ApplyImageDataFilters(TestData.GetImageData(), ImageTypes.Wcf)
                .Where(descr => desc.Verion == "3.5");
```

tests/Microsoft.DotNet.Framework.Docker.Tests/SdkOnlyImageTests.cs
mthalman: Yeah, forgot to mention that in the initial comments. The need to use skipping is because XUnit doesn't allow a theory's MemberData method to return an empty set. So we can't filter out the member data; the test needs to skip the stuff that it doesn't want instead. Same applies to your other comment.

MichaelSimons: Oh right, I remember encountering this before.  That said, I still see value in expressing this via logic versus a static list.  The value is in maintainability as versions reach EOL (ha ha).

```
imageDescriptor.Version != imageDescriptor.SdkVersion;
```

mthalman:
I made some changes to handle a case where it was still running into the XUnit issue of having an empty set for the member data. That can happen when the paths filter causes the image for a given image type to be filtered out. For example, a 3.5 test leg would have the following paths specified:
* src/runtime/3.5/windowsservercore-ltsc2019
* src/sdk/3.5/windowsservercore-ltsc2019
* src/aspnet/3.5/windowsservercore-ltsc2019

There is no wcf path here, by design, so this causes the WCF tests to end up with an empty set of image data and XUnit throws. To work around this, I've added code to compensate for this scenario by injecting a null `ImageDescriptor` as a placeholder to allow the test to detect and skip execution in that case. I've also added some validation logic to ensure we only end up in that state for this expected scenario. This also applies to the SDK-only tests.

tests/Microsoft.DotNet.Framework.Docker.Tests/SdkOnlyImageTests.cs
```diff
@@ -22,14 +22,28 @@ public SdkOnlyImageTests(ITestOutputHelper outputHelper)
 
         protected override string ImageType => "sdk";
 
-        public static IEnumerable<object[]> GetImageData() =>
-            ImageTestHelper.ApplyImageDataFilters(TestData.GetImageData(), "sdk");
+        public static IEnumerable<object[]> GetImageData()
+        {
+            IEnumerable<object[]> imageData =
+                ImageTestHelper.ApplyImageDataFilters(TestData.GetImageData(), ImageTypes.Sdk);
+            if (!imageData.Any())
```
MichaelSimons, line 29: Would it make sense to move this into the `ImageTestHelper.ApplyImageDataFilters` method? This would eliminate the duplication between the SDK and WCF tests.  If you think it would be harmful to have for the other test classes, you could condition the logic on an optional parameter - e.g. "allowEmptyResults"

MichaelSimons (approved):

PR #1036, opened 2022-12-19
Remove .NET Core 3.1 from SDK Dockerfiles

### LINKED ISSUE #994: Remove .NET Core 3.1 runtime from SDK when unsupported
.NET Core 3.1 will be EOL on Dec 13, 2022. At that time, .NET Core 3.1 runtime component should be removed as part of the installation of VS Build Tools in the SDK images.

https://github.com/microsoft/dotnet-framework-docker/blob/3638c1811c7aa316daeb4f6c143db22d10c84256/eng/dockerfile-templates/sdk/Dockerfile#L55

Related to https://github.com/microsoft/dotnet-framework-docker/issues/984

mthalman:
.NET Core 3.1 is EOL as of Dec. 13, 2022 so it is being removed from the SDK Dockerfiles.

Fixes #994


MichaelSimons (approved):
Am I correct in that this won't be released until January servicing?

PR #1087, opened 2023-08-08
Updates for .NET Fx servicing August 2023

lbussell:
All of the `ndp48/ndp481` extensions got removed from the cab files this time around so there were some necessary template changes.

### DIFF
--- a/manifest.versions.json
+++ b/manifest.versions.json
@@ -1,11 +1,11 @@
 {
     "variables": {
-      "kb|ltsc2016|3.5": "kb5028169",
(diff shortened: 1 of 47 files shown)


manifest.versions.json
mthalman, line 1: VS version should be updated to 17.7.

manifest.versions.json
```diff
@@ -36,7 +36,7 @@
 
       "nuget|version": "6.6.1",
 
-      "vs|version": "17.6",
+      "vs|version": "17.7",
```
mthalman, line 39: Dockerfiles need to be regen'd.

lbussell: The regen doesn't do anything, the ".6"/".7" part of the version doesn't show up in any Dockerfiles.

mthalman: You're right. My bad. It's just used for tests.

mthalman (approved):

PR #1091, opened 2023-09-12
Update templates to reflect 48/481 suffix in cab files

lbussell:


MichaelSimons (approved):

mthalman (commented):
I'm late but looks good.

PR #1101, opened 2023-12-12
Updates for December 2023 Patch Tuesday

lbussell:


manifest.datestamps.json
```diff
@@ -3,35 +3,39 @@
       "RuntimeReleaseDateStamp": "20231114",
       "AspnetReleaseDateStamp": "20231114",
       "WcfReleaseDateStamp": "20231114",
-      "SdkReleaseDateStamp": "20231114",
+      "SdkReleaseDateStamp": "20231211",
```
mthalman, line 6: Ideally this would be `20231212` to avoid confusion.

mthalman (approved):

PR #1110, opened 2024-03-11
Updates for Windows servicing (March 2024)

lbussell:


manifest.versions.json
```diff
@@ -34,7 +34,7 @@
       "kb|ltsc2019|default": "$(kb|ltsc2019|4.7.2)",
       "lcu|ltsc2019|default": "$(lcu|ltsc2019|4.7.2)",
 
-      "nuget|version": "6.8.0",
+      "nuget|version": "6.9.1",
```
mthalman, line 37: Because of this change, all the sdk Dockerfiles will be changed. Which means we'll need the datestamp applied to tags for all of them.

mthalman (approved):

PR #1127, opened 2024-05-21
Update VS to 17.10

lbussell:


mthalman (approved):

lbussell:
Test failures must be due to some caching somewhere - built locally and vswhere picks up 17.10 no problem.

mthalman:
Yeah, I just downloaded the build agent from the URL we use and it had the new version. Did the build run before VS released?

lbussell:
> Yeah, I just downloaded the build agent from the URL we use and it had the new version. Did the build run before VS released?

No, I checked the test agent locally before I ever submitted the PR.

PR #1136, opened 2024-06-06
Suppress CodeQL alerts for generated code

mthalman:
Adds CodeQL suppressions for some issues that are being alerted upon from the aspnetapp sample. This code comes from the project template. I checked with the latest templates from Visual Studio and they have the same code as we have currently.


lbussell (approved):

lbussell:
Theoretically .NET Framework is getting these same alerts in its source code. Should we file a follow-up issue to remove this if that is ever fixed? It may also be unlikely to be fixed.

mthalman:
> Theoretically .NET Framework is getting these same alerts in its source code. Should we file a follow-up issue to remove this if that is ever fixed? It may also be unlikely to be fixed.

That would happen naturally whenever we update samples. We'd just copy over our existing files with the new content.

PR #1161, opened 2024-08-19
Switch markdownlint GH action to use super-linter

MichaelSimons:
Related to https://github.com/dotnet/dotnet-docker/issues/5316

Switched the MarkDown linter GH action to use the super-linter. This gives us the advantage of enabling other linters via a "flip of a switch". The configuration will be more standardized with the super-linter versus having separate GH actions/linters. The super-linter used the markdownlint tool we are already using so there should be no functional changes.


MichaelSimons:
A test environment with these changes can be seen [here](https://github.com/MichaelSimons/dotnet-framework-docker/pulls).  I will open a draft test PR after merging as a validation check the ensure the E2E is working as expected.

Issue #1117, opened 2024-03-20
`RuntimeSdkImageTests.VerifyImagesWithWebApps` test is failing in PR validation on `sdk:4.8-windowsservercore-ltsc2016` image


lbussell:
WindowsLtsc2016_amd64 4.8-windowsservercore-ltsc2016-runtime tests are failing consistently in PR validation. This isn't happening in the internal CI.

Link to failing PR validation run: https://dev.azure.com/dnceng-public/public/_build/results?buildId=608811&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=b4a7524f-24ff-5cd3-3045-5ce9862a6c27

Test failure outputs (multiple across different tests):
```
System.InvalidOperationException : Failed to execute docker run --name webapp-133553521378571687 --rm  webapp-133553521378571687 powershell -command "dir ./bin/SimpleWebApplication.dll"
Exit code: 125
Standard Error: docker: Error response from daemon: container bed4570c99656a9e43c0f076450fcc1a9fd9826b45eb85f3a5e569fe1a61da74 encountered an error during hcsshim::System::Start: failure in a Windows system call: This operation returned because the timeout period expired. (0x5b4).
```

```
System.InvalidOperationException : Failed to execute docker build -t webapp-133553523877594066 --build-arg BASE_BUILD_IMAGE=mcr.microsoft.com/dotnet/framework/sdk:4.8-windowsservercore-ltsc2016 -f D:\a_work\1\a\bin\Microsoft.DotNet.Framework.Docker.Tests\release\projects\webapp-4.8\Dockerfile D:\a_work\1\a\bin\Microsoft.DotNet.Framework.Docker.Tests\release\projects\webapp-4.8
Exit code: -1
Standard Error: The command 'powershell -Command $ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue'; msbuild SimpleWebApplication.sln /p:Configuration=Release' returned a non-zero code: 4294967295: failed to shutdown container: container 51f610b096f674c819bd3642b8b8bd8b3dbb8f2f87c7843709dc939b3199096b encountered an error during hcsshim::System::Shutdown: failure in a Windows system call: The handle is invalid. (0x6)
```

lbussell:
This is now failing to start the build stage with this awesome stack trace:

```
[...]

PR #1175, opened 2024-09-30
Fix MAR search links with trailing slash

lbussell:
The MAR link https://mcr.microsoft.com/catalog?search=dotnet/framework/ results in a page that says "Sorry, we encountered an error. Please try again later". Removing the trailing slash fixes it, so I've adjusted the templates accordingly.

image


mthalman (approved):
This is fine as a temporary measure.

I've filed an issue for MAR because this should get fixed: https://github.com/microsoft/containerregistry/issues/171. We really do want the `dotnet/` query to work properly to avoid getting hits on other teams' .NET-related repos but not within the scope of the `dotnet/` repo.

lbussell:
> I've filed an issue for MAR...

Thanks, I threw that issue in the tracking column of the .NET Docker board.

lbussell:
I fixed other broken links that were upsetting the link checker - the same as https://github.com/dotnet/dotnet-docker/pull/5923/commits/3562f21dc388f9dc07e4e2aab5351a1d24b813a5

lbussell:
Build failure is https://github.com/microsoft/dotnet-framework-docker/issues/1117, merging around that since it's unrelated to changes.

PR #1194, opened 2024-12-09
Windows Server Core LTSC 2016 Updates for December 2024

lbussell:
There is a new version of NuGet CLI available (6.12.1), but we should probably hold off until January to apply that update to all images.


manifest.datestamps.json
```diff
@@ -1,40 +1,45 @@
...
       "4.7.2-ltsc2019-Aspnet-DateStamp": "$(AspnetReleaseDateStamp)",
       "4.7.2-ltsc2019-Wcf-DateStamp": "$(WcfReleaseDateStamp)",
-      "4.8-ltsc2016-Runtime-DateStamp": "$(RuntimeReleaseDateStamp)",
-      "4.8-ltsc2016-Aspnet-DateStamp": "$(AspnetReleaseDateStamp)",
-      "4.8-ltsc2016-Wcf-DateStamp": "$(WcfReleaseDateStamp)",
-      "4.8-ltsc2016-Sdk-DateStamp": "$(SdkReleaseDateStamp)",
+      "4.8-ltsc2016-Runtime-DateStamp": "$(ltsc2016-RuntimeReleaseDateStamp)",
+      "4.8-ltsc2016-Aspnet-DateStamp": "$(ltsc2016-AspnetReleaseDateStamp)",
+      "4.8-ltsc2016-Wcf-DateStamp": "$(ltsc2016-WcfReleaseDateStamp)",
+      "4.8-ltsc2016-Sdk-DateStamp": "$(ltsc2016-SdkReleaseDateStamp)",
```
mthalman, line 42: This should be reverted. We don't install the new KB in the 4.8 Dockerfiles.

lbussell: I updated it such that only 3.5 LTSC 2016 images have the new datestamp.

mthalman: Now it's not updating enough. There are still other non-3.5 versions which are being updated to reference the new KB number. All of those versions should have an updated datestamp.

lbussell: OK - Tried to make sure it matches the Dockerflies that were changed this time. I added 4.7, 4.7.1, and 4.7.2.

PR #1197, opened 2024-12-17
Update ASP.NET samples with latest templates

lbussell:
- Updated `aspnetapp` sample to latest WebForms template.
- Updated `aspnetmvcapp` sample to latest MVC template.


lbussell:
Confirmed `aspnetapp` image builds and runs correctly.

The `aspnetmvcapp` image stops immediately with the message `Stopping service 'w3svc'`.

samples/aspnetapp/aspnetapp/Scripts/WebForms/Menu.js
```diff
@@ -170,7 +170,6 @@ function Menu_GetData(item) {
         var a = (item.tagName.toLowerCase() == "a" ? item : WebForm_GetElementByTagName(item, "a"));
         var menu = Menu_FindMenu(a);
         try {
-            // CodeQL [SM04509] Code generated from project template
```
mthalman, line 173: It looks like these changes from https://github.com/microsoft/dotnet-framework-docker/pull/1136 were removed. These should be added back since the underlying issues aren't fixed.

samples/aspnetapp/aspnetapp/Scripts/WebForms/MSAjax/MicrosoftAjaxWebForms.js
```diff
@@ -3,5 +3,4 @@
 // Copyright (C) Microsoft Corporation. All rights reserved.
 //----------------------------------------------------------
 // MicrosoftAjaxWebForms.js
-// CodeQL [SM01623] Code generated from project template
```
mthalman, line 6: This should be added back too.

mthalman (approved):

Issue #1213, opened 2025-02-13
Update to msbuild 17.13 breaks builds


kewdgr:
### Describe the Bug

All our builds have been working for years with tags up to 3.5-20250114-windowsservercore-ltsc2019. 
It seems that the switch to VS Build Tools 17.13 introduced a breaking change for our builds

message: Exception has been thrown by the target of an invocation.  Could not load file or assembly 'Microsoft.VisualStudio.SolutionPersistence, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.

Does anybody have a suggestion for a fix? A Tag for a container using older msbuild? 

### Steps to Reproduce

Just starting MSBuild included in this container 3.5-20250211-windowsservercore-ltsc2019 was enough.
It works perfectly with the same project in this container : 3.5-20250114-windowsservercore-ltsc2019

### Other Information

message: Exception has been thrown by the target of an invocation.  Could not load file or assembly 'Microsoft.VisualStudio.SolutionPersistence, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.

### Output of `docker version`

Server: Docker Desktop 4.38.0 (181591)
 Engine:
  Version:          27.5.1
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.11
  Git commit:       4c9b3b0
  Built:            Wed Jan 22 13:40:41 2025
  OS/Arch:          windows/amd64
  Experimental:     false

### Output of `docker info`

Client:
 Version:    27.5.1
 Context:    desktop-windows
 Debug Mode: false
 Plugins:
  ai: Ask Gordon - Docker Agent (Docker Inc.)
    Version:  v0.7.3
    Path:     C:\Users\user\.docker\cli-plugins\docker-ai.exe
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.20.1-desktop.2
    Path:     C:\Users\user\.docker\cli-plugins\docker-buildx.exe
  compose: Docker Compose (Docker Inc.)
    Version:  v2.32.4-desktop.1
    Path:     C:\Users\user\.docker\cli-plugins\docker-compose.exe
  debug: Get a shell into any image or container (Docker Inc.)
    Version:  0.0.38
    Path:     C:\Users\user\.docker\cli-plugins\docker-debug.exe
[...]

PR #1214, opened 2025-02-13
Wrap net35 app in a solution

rainersigwald:
To see if this is a simple repro of #1213.


MichaelSimons:
@lbussell, @mthalman  - Did the scenario tests actually run here?  It doesn't look like it.  Image caching?  Does a Dockerfile need to be touched?

```
[xUnit.net 00:02:36.73]     Microsoft.DotNet.Framework.Docker.Tests.WcfImageTests.VerifyImagesWithApps(imageDescriptor: null) [SKIP]
```

mthalman:
Yes, it ran them. It ran 20 tests and the log shows them. The WCF ones are skipped because there are now 3.5 WCF Dockerfiles.

PR #1196, opened 2024-12-11
Remove .NET 6 from SDK images

lbussell:
See https://github.com/dotnet/dotnet-docker/issues/6051#issuecomment-2474763943

PR #1216, opened 2025-02-27
Remove .NET 6 SDK from Windows Server 2016 SDK images

lbussell:
These were left out from https://github.com/microsoft/dotnet-framework-docker/pull/1196.

This change was already announced in https://github.com/microsoft/dotnet-framework-docker/discussions/1203.

I took the opportunity to factor out the VS buildtools installation to another template so that this is less likely to happen in the future.


mthalman (approved):

lbussell:
This commit was included in https://github.com/microsoft/dotnet-framework-docker/pull/1225

PR #1217, opened 2025-02-28
Use combined reference assembly/targeting pack zip file

### LINKED ISSUE #1135: Improve targeting pack acquisition
The pattern we use for acquiring targeting packs no longer makes sense. We chose the pattern because we were still creating new .NET Framework versions every year or two and new targeting packs with them. We needed a versioning scheme where the target framework archives we publish didn't change (for compat). We designed a scheme for that, which can be seen in the implementation. In retrospect, we could have chosen a simpler scheme, but that's now history.

https://github.com/microsoft/dotnet-framework-docker/blob/58a18cf958883c6a67ae888854bd9bb09f085e83/src/sdk/4.8.1/windowsservercore-ltsc2022/Dockerfile#L71-L83

Let's assume the following:

- No new .NET Framework versions and therefore no new target frameworks
- Content integrity is a key tenet of delivering high-quality software
- The pattern above is cryptic and not all that attractive.

We should then:

- Make a new target framework archive, including all the target frameworks. We can still version it in some way (like timestamp it).
- Create a content hash for it validate that in the Dockerfile (like we do in other Dockerfiles).
- Acquire this single archive in a simpler more standard way (no funky looping).

lbussell:
Fixes #1135 

This PR is marked as draft because we still need to publish the reference assembly zip to our storage account.

Here's the PowerShell script that I used to generate the new reference assembly zip:

<details>

```pwsh
$blobs = @(
    "v4.0.zip",
    "v4.5.2.zip",
    "v4.6.2.zip",
    # v4.7.1 is not used and conflicts with 4.7.2's files
    # "v4.7.1.zip",
    "v4.7.2.zip",
    "v4.8.1.zip",
    "v4.8.zip"
)

foreach ($blobname in $blobs) {
    $url = "https://dotnetbinaries.blob.core.windows.net/referenceassemblies/$blobname"
    $output = "./$blobname"
    Invoke-WebRequest -Uri $url -OutFile $output
}

foreach ($blobname in $blobs) {
    Expand-Archive -path "./$blobname" -DestinationPath "./temp" -Force
}

Compress-Archive -Path "./temp/*" -DestinationPath "./referenceassemblies.v0.zip"
```

</details>

### DIFF
--- a/eng/dockerfile-templates/sdk/Dockerfile
+++ b/eng/dockerfile-templates/sdk/Dockerfile
@@ -91,19 +91,7 @@ RUN powershell setx /M PATH $(${Env:PATH} `
     + \";${Env:ProgramFiles(x86)}\Microsoft SDKs\Windows\v10.0A\bin\NETFX {{sdkVersion}} Tools\" `
     + \";${Env:ProgramFiles(x86)}\Microsoft SDKs\ClickOnce\SignTool\")
 
-# Install Targeting Packs
-RUN powershell " `
-    $ErrorActionPreference = 'Stop'; `
-    $ProgressPreference = 'SilentlyContinue'; `
-    @@('4.0', '4.5.2', '4.6.2', '4.7.2', '4.8', '4.8.1') `
-    | %{ `
-        Invoke-WebRequest `
-            -UseBasicParsing `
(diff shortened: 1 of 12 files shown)


lbussell:
Closing in favor of https://github.com/microsoft/dotnet-framework-docker/issues/1221

PR #1239, opened 2025-04-11
Apply 4B updates

mthalman:


mthalman:
Looks good.

lbussell:
There is a new KB available for Server 2016, [kb5058921](https://support.microsoft.com/en-us/topic/april-11-2025-kb5058921-os-build-14393-7973-out-of-band-bcd6c4ef-eb5e-43e1-9a6e-03e7113ff30a). However it is not a security fix and does not contain any fixes for containers. Therefore I don't think we should pick it up because we don't typically install OOB non-security updates. So the PR is good as-is, IMO.

@mthalman, thoughts?

lbussell:
I also confirmed that the base 2016 image doesn't contain any OOB updates

```
> docker run --rm mcr.microsoft.com/windows/servercore:ltsc2016-amd64 powershell -Command get-hotfix

Source        Description      HotFixID      InstalledBy          InstalledOn
------        -----------      --------      -----------          -----------
24CB7BBD36B3  Update           KB4054590                          4/5/2025 1...
24CB7BBD36B3  Update           KB4132216                          4/5/2025 1...
24CB7BBD36B3  Security Update  KB5055661                          4/5/2025 1...
24CB7BBD36B3  Security Update  KB5055521                          4/5/2025 1...
```

mthalman:
Yeah, I agree we can skip that.

jander-msft (approved):

